PR #42, opened 2018-11-23
Add docs for Job resource and update upgrade docs

darkowlzz:


README.md
```diff
@@ -132,9 +132,11 @@ of StorageOS.
 An example of of `StorageOSUpgrade` resource is [storageos_v1alpha1_storageosupgrade_cr.yaml](/deploy/crds/storageos_v1alpha1_storageosupgrade_cr.yaml).
 
 Only offline upgrade is supported for now by cluster-operator. During the
-upgrade, the applications that use StorageOS volumes are scaled down and the
-whole StorageOS cluster is restarted with a new version. Once the StorageOS
-cluster becomes usable, the applications are scaled up to their previous configuration.
+upgrade, StorageOS maintenance mode is enabled, the applications that use
+StorageOS volumes are scaled down and the whole StorageOS cluster is restarted
+with a new version. Once the StorageOS cluster becomes usable, the applications
+are scaled up to their previous configuration. The maintenance mode can be
```
Arau: "The maintenance mode can be disabled" Or shouldn't it say: "Once the upgrade is complete, make sure you delete the upgrade resource to put the StorageOS cluster in normal mode rather than in maintenance mode"

or something alike?

README.md
```diff
@@ -132,9 +132,11 @@ of StorageOS.
 An example of of `StorageOSUpgrade` resource is [storageos_v1alpha1_storageosupgrade_cr.yaml](/deploy/crds/storageos_v1alpha1_storageosupgrade_cr.yaml).
 
 Only offline upgrade is supported for now by cluster-operator. During the
-upgrade, the applications that use StorageOS volumes are scaled down and the
-whole StorageOS cluster is restarted with a new version. Once the StorageOS
-cluster becomes usable, the applications are scaled up to their previous configuration.
+upgrade, StorageOS maintenance mode is enabled, the applications that use
```
Arau: I think it is not clear that the operator will stop your apps. I think if we do scale down their deployments, the user must be warned energetically.

README.md
```diff
@@ -132,9 +132,11 @@ of StorageOS.
...
+StorageOS volumes are scaled down and the whole StorageOS cluster is restarted
+with a new version. Once the StorageOS cluster becomes usable, the applications
+are scaled up to their previous configuration. The maintenance mode can be
+disabled by deleting the upgrade resource.
 
 Once an upgrade resource is created, events related to the upgrade can be
 viewed in the upgrade object description. All the status and errors, if any,
```
Arau: I think is incomplete:

That you can see by `kubectl -n storageos-operator describe XXX`

Arau: or `kubectl -n storageos-operator get event`, etc

README.md
```diff
@@ -167,6 +172,120 @@ Parameter | Description | Default
 `newImage` | StorageOS node container image to upgrade to |
 
 
+## Cleanup Old Configurations
+
+StorageOS creates and saves its files at `/var/lib/storageos` on the hosts. This
```
Arau: We need to emphasise that this will remove all their data and it will be unrecoverable (big warning). We should also mention that in case of using external etcd, they should remove all the data in the storageos section. 
something like 
`ETCDCTL_API=3 /usr/local/bin/etcdctl --endpoints http://storageos-etcd-client:2379 del --prefix storageos`

PR #59, opened 2019-02-17
Add daemonset node toleration support

### LINKED ISSUE #58: Add daemonset node tolerations support
We've the desire to deploy specific storage nodes. In order to achieve this we set a node taint to avoid scheduling any workload pods onto them. In order to schedule the storageos pods onto them, the operator needs to inject a node toleration in the daemonset. I would like to see this be supported by adding support in similar way as #19 did for node selectors.

What are your thoughts about this?
If you're willing to support this, I could find some time to workout a draft PR based on the code of #19.

tobiaskohlbau:
Hello, I've started implementing support for tolerations within the operator. I would like to get early feedback if I'm on the right track. Plase be patient with me as it's the first time fiddling arround with kubernetes internals for me 😊

Fixes #58

### DIFF
--- a/pkg/apis/storageos/v1alpha1/job_types.go
+++ b/pkg/apis/storageos/v1alpha1/job_types.go
@@ -35,6 +35,10 @@ type JobSpec struct {
 	// NodeSelectorTerms is the set of placement of the job pods using node
 	// affinity requiredDuringSchedulingIgnoredDuringExecution.
 	NodeSelectorTerms []corev1.NodeSelectorTerm `json:"nodeSelectorTerms"`
+
+	// Tolerations is to set the placement of storageos pods using
+	// pod toleration.
+	Tolerations []corev1.Toleration `json:"tolerations"`
 }
 
 // GetLabelSelector returns Job's pod label selector.
--- a/pkg/controller/job/job_controller.go
+++ b/pkg/controller/job/job_controller.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"strings"
 	"time"
+	"fmt"
 
 	storageosv1alpha1 "github.com/storageos/cluster-operator/pkg/apis/storageos/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
@@ -280,6 +281,17 @@ func newDaemonSetForCR(cr *storageosv1alpha1.Job) (*appsv1.DaemonSet, error) {
 		},
 	}
 
+	// Add pod tolerations if defined.
+	tolerations := cr.Spec.Tolerations
+	for i := range tolerations {
+		if tolerations[i].Operator == corev1.TolerationOpExists && tolerations[i].Value != "" {
+			return nil, fmt.Errorf("key(%s): toleration value must be empty when `operator` is 'Exists'", tolerations[i].Key)
+		}
+	}
+	if len(tolerations) > 0 {
+		dset.Spec.Template.Spec.Tolerations = cr.Spec.Tolerations
+	}
+	
 	// Add node affinity if defined.
 	if len(cr.Spec.NodeSelectorTerms) > 0 {
 		dset.Spec.Template.Spec.Affinity = &corev1.Affinity{NodeAffinity: &corev1.NodeAffinity{
--- a/pkg/storageos/deploy.go
+++ b/pkg/storageos/deploy.go
@@ -9,7 +9,7 @@ import (
 	"github.com/blang/semver"
 	api "github.com/storageos/cluster-operator/pkg/apis/storageos/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
@@ -229,7 +229,7 @@ func (s *Deployment) Deploy() error {
 }
 
 func (s *Deployment) createNamespace() error {
-	ns := &v1.Namespace{
+	ns := &corev1.Namespace{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "v1",
 			Kind:       "Namespace",
@@ -247,7 +247,7 @@ func (s *Deployment) createNamespace() error {
 }
 
 func (s *Deployment) createServiceAccount(name string) error {
-	sa := &v1.ServiceAccount{
+	sa := &corev1.ServiceAccount{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "v1",
 			Kind:       "ServiceAccount",
@@ -524,7 +524,7 @@ func (s *Deployment) createClusterRoleBindingForAttacher() error {
 func (s *Deployment) createDaemonSet() error {
 	ls := labelsForDaemonSet(s.stos.Name)
 	privileged := true
-	mountPropagationBidirectional := v1.MountPropagationBidirectional
+	mountPropagationBidirectional := corev1.MountPropagationBidirectional
 	allowPrivilegeEscalation := true
 
 	dset := &appsv1.DaemonSet{
@@ -543,20 +543,20 @@ func (s *Deployment) createDaemonSet() error {
 			Selector: &metav1.LabelSelector{
 				MatchLabels: ls,
 			},
-			Template: v1.PodTemplateSpec{
+			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: ls,
 				},
-				Spec: v1.PodSpec{
+				Spec: corev1.PodSpec{
 					ServiceAccountName: "storageos-daemonset-sa",
 					HostPID:            true,
 					HostNetwork:        true,
-					DNSPolicy:          v1.DNSClusterFirstWithHostNet,
-					InitContainers: []v1.Container{
+					DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
+					InitContainers: []corev1.Container{
 						{
 							Name:  "enable-lio",
 							Image: s.stos.Spec.GetInitContainerImage(),
-							VolumeMounts: []v1.VolumeMount{
+							VolumeMounts: []corev1.VolumeMount{
 								{
 									Name:      "kernel-modules",
 									MountPath: "/lib/modules",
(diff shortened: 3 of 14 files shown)


pkg/apis/storageos/v1alpha1/job_types.go
```diff
@@ -35,6 +35,10 @@ type JobSpec struct {
 	// NodeSelectorTerms is the set of placement of the job pods using node
 	// affinity requiredDuringSchedulingIgnoredDuringExecution.
 	NodeSelectorTerms []corev1.NodeSelectorTerm `json:"nodeSelectorTerms"`
+
+	// Tolerations is to set the placement of storageos pods using
```
tobiaskohlbau: I would like to receive comments if the job needs to be toleration aware too? I did not find the same content in the `job` CRD as within `storageoscluster`.

pkg/storageos/deploy.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/blang/semver"
 	api "github.com/storageos/cluster-operator/pkg/apis/storageos/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
tobiaskohlbau: Looks like `gofmt` wants to add this v1. I hope this is fine.

Arau:
Adding @darkowlzz as a reviewer.

pkg/storageos/deploy.go
```diff
@@ -1087,6 +1089,14 @@ func (s *Deployment) addNodeAffinity(podSpec *v1.PodSpec) {
 	}
 }
 
+// addToleration adds toleration to the given pod spec from the cluter
+// spec Toleration.
```
darkowlzz: Keyword typos.
`Toleration` -> `Tolerations`.
It's incorrect in `addNodeAffinity` comment as well. Should be `NodeSelectorTerms`, not `NodeSelectorLabel`.

pkg/storageos/deploy_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"testing"
 
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
darkowlzz: Let's rename it to `corev1`.

pkg/storageos/deploy.go
```diff
@@ -1087,6 +1089,14 @@ func (s *Deployment) addNodeAffinity(podSpec *v1.PodSpec) {
...
+func (s *Deployment) addToleration(podSpec *v1.PodSpec) {
+	if len(s.stos.Spec.Tolerations) > 0 {
+		podSpec.Tolerations = s.stos.Spec.Tolerations
+	}
```
darkowlzz: I think we should add a check here to ensure that `Toleration.Value` is not set if `Toleration.Operator` is `TolerationOpExists`. Passing a value with exists makes the configuration invalid. This would make this function to return an error. An error check should be added to the `addToleration()` call in `createDaemonSet()`. With that, you be able to test this by adding a test case in the `TestDeployTolerations` in deploy_test.go where you set value with operator exists. Deployment.Deploy() would return an error.

Before adding this, can you check if adding the invalid configuration fails at `createOrUpdateObject()` call at the end of `createDaemonSet()`. If it does, and a proper error is visible in the description of the storageoscluster object events, maybe we need not add this check.

pkg/apis/storageos/v1alpha1/job_types.go
darkowlzz: You should also add tolerations in the job controller, similar to this https://github.com/storageos/cluster-operator/blob/v0.0.8/pkg/controller/job/job_controller.go#L283-L290.

pkg/storageos/deploy.go
tobiaskohlbau: Looks like this is a bit of documentation lack from kubernetes site. As in the documentation is stated:
> the operator is Exists (in which case no value should be specified)

and the sourcecode show that is should work.

https://github.com/kubernetes/kubernetes/blob/a9fd9cef765a0b6d3854b187dcb6d05097f50c6a/staging/src/k8s.io/api/core/v1/toleration.go#L37-L56

But I found that the validation checks this:

https://github.com/kubernetes/kubernetes/blob/f7c4389b793cd6cf0de8d67f2c5db28b3985ad59/pkg/apis/core/validation/validation.go#L2813-L2855

I've tried to run the tests with `TolerationOpExists` and it does not fail, therefore I will implement the check early on.

tobiaskohlbau:
I've updated my PR and hope I resolved all addressed reviews.

pkg/storageos/deploy_test.go
```diff
@@ -815,6 +815,113 @@ func TestDeployNodeAffinity(t *testing.T) {
...
+					Operator: corev1.TolerationOpEqual,
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+		},
+	}
+
+	c := fake.NewFakeClient()
+	deploy := NewDeployment(c, stosCluster, nil, testScheme, "", false)
+	deploy.Deploy()
```
darkowlzz: Can we combine the two separate test sections to one?
This `deploy.Deploy()` will internally call `addTolerations()`. Hence, `Deploy()` will return error when the toleration configuration is invalid. Add an error check here.
Let's create the StorageOSCluster object inside a test run, and set the toleration from the test cases, like in the first test section. With that, you can have exists, equal, with value and without value, all covered in one set of test cases.
Only if there's no error and toleration is specified, you compare the toleration of the deployed pod and the cluster spec.

pkg/controller/job/job_controller.go
```diff
@@ -280,6 +280,11 @@ func newDaemonSetForCR(cr *storageosv1alpha1.Job) (*appsv1.DaemonSet, error) {
 		},
 	}
 
+	// Add pod tolerations if defined.
+	if len(cr.Spec.Tolerations) > 0 {
+		dset.Spec.Template.Spec.Tolerations = cr.Spec.Tolerations
```
darkowlzz: We need the toleration exists and value check here as well.

pkg/storageos/deploy.go
```diff
@@ -766,11 +771,11 @@ func (s *Deployment) addNodeContainerResources(nodeContainer *v1.Container) {
 // KubeletPluginsWatcher. This is used to change the CSI driver registry setup
 // based on the kubernetes cluster setup.
 func kubeletPluginsWatcherSupported(version string) bool {
-	// Supported if v1.12.0 or above.
+	// Supported if corev1.12.0 or above.
```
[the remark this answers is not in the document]
tobiaskohlbau: sorry I think my regex skills are not strong with me 😅

pkg/storageos/deploy.go
```diff
@@ -1568,7 +1588,7 @@ func (s *Deployment) createStorageClass() error {
 	if s.stos.Spec.CSI.Enable {
 		// Add CSI creds secrets in parameters.
 		if CSIV1Supported(s.k8sVersion) {
-			// New CSI secret parameter keys were introduced in CSI v1.
+			// New CSI secret parameter keys were introduced in CSI corev1.
```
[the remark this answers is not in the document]
tobiaskohlbau: sorry I think my regex skills are not strong with me 😅

pkg/storageos/deploy_test.go
tobiaskohlbau: I've tried using deploy.Deploy but it throws errors even when using the correct handling. But I will give it another shot.

darkowlzz:
It's good that the e2e tests were not affected by this change.
Would you like to add e2e test changes for this as well?
It can be done separately, but if you would like to give it a try, you should first enable taint on the single node somewhere here https://github.com/storageos/cluster-operator/blob/master/test/e2e.sh#L202.
I think the node name includes the travis CI job ID, which should be $TRAVIS_JOB_ID as per https://docs.travis-ci.com/user/environment-variables/#default-environment-variables.
Or even better, you can use kubectl to get it with something like
```
$ kubectl get nodes --no-headers=true -o=name
```

And then, set the toleration in the cluster spec https://github.com/storageos/cluster-operator/blob/master/test/e2e/clusterCSI_test.go#L27-L34 and https://github.com/storageos/cluster-operator/blob/master/test/e2e/clusterInTreePlugin_test.go#L25-L29.

Also, I think the taint won't allow the operator pod to be scheduled, because there's just one node. So, you might have to add a taint in the operator deployment as well https://github.com/storageos/cluster-operator/blob/master/deploy/operator.yaml.

That would make it work in the e2e test.

If not, it's fine, we can add it separately.

darkowlzz:
While thinking about the e2e test scenario, I realized that it'll be better to add the toleration to the statefulsets as well. Else, CSI deployment will fail if there's just one node, because the helper containers in the statefulsets will not be scheduled. Let's add it to statefulset as well.
You can get podSpec similar to the createDaemonSet() function in createStatefulSet() function, somewhere here https://github.com/storageos/cluster-operator/blob/master/pkg/storageos/deploy.go#L1231, and call the same `addTolerations()` function. That should be enough for it to work.
With this, the e2e test will not fail for CSI deployment.

pkg/controller/job/job_controller.go
tobiaskohlbau: You're totally right, my bad. Sorry for so much inconvenience.

pkg/storageos/deploy_test.go
darkowlzz: If the error you get is something like this https://github.com/storageos/cluster-operator/blob/master/pkg/controller/storageoscluster/storageoscluster_controller.go#L224-L232 , you can add a retry. While running through the controller, the controller runs the deployment again in order to ensure all the resources have been deployed.
If it's something else and you're not sure what to do with it, please share the error. We can figure out what to do with it.

tobiaskohlbau: I've tried to implement your suggestions. I'm now back to the error:

`expected no error but got one: no kind is registered for the type v1alpha1.StorageOSCluster in scheme "k8s.io/client-go/kubernetes/scheme/register.go:60"`

I think the problem is that the test does not register the scheme and therefore underlying functions fail. As storageos does not wrap errors within `deploy.Deploy()`  I'm not able to tell which inner function is causing this.

tobiaskohlbau:
I would like to add in advance that I've taken the changes and deployed them onto our cluster and everything worked like expected.

pkg/storageos/deploy_test.go
```diff
@@ -861,6 +861,114 @@ func TestDeployNodeAffinity(t *testing.T) {
...
+
+			if err := c.Get(context.Background(), nsName, createdDaemonset); err != nil {
+				t.Fatal("failed to get the created daemonset", err)
+			}
+
+			podSpec := createdDaemonset.Spec.Template.Spec
+
+			if !reflect.DeepEqual(podSpec.Tolerations, stosCluster.Spec.Tolerations) {
+				t.Errorf("unexpected Tolerations value:\n\t(GOT) %v\n\t(WNT) %v", podSpec.Tolerations, stosCluster.Spec.Tolerations)
+			}
```
darkowlzz: We should check the same for the statefulset too, in case of CSI only.
I'll add that in a follow-up PR.

PR #67, opened 2019-03-01
Update description and move to stable

croomes:
Add more product info and move from alpha to stable channel.

### DIFF
--- a/deploy/olm-operators.configmap.yaml
+++ b/deploy/olm-operators.configmap.yaml
@@ -292,11 +292,11 @@ data:
         minKubeVersion: "1.10.0"
         maturity: alpha
         maintainers:
-        - name: StorageOS, Ltd
+        - name: StorageOS, Inc
           email: support@storageos.com
 
         provider:
-          name: StorageOS, Ltd
+          name: StorageOS, Inc
 
         labels:
           alm-status-descriptors: storageosoperator.v0.0.9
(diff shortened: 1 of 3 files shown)


deploy/stos.subscription.yaml
```diff
@@ -7,4 +7,4 @@ spec:
   source: olm-operators
   sourceNamespace: olm
   name: storageos
-  channel: alpha
+  channel: stable
```
darkowlzz: This is used in the e2e test setup. And changing this without changing https://github.com/storageos/cluster-operator/blob/master/deploy/olm-operators.configmap.yaml#L471 will fail the operator installation. But I'm working on changing the configmap a bit, so let's remove this change and just change the channel in https://github.com/storageos/cluster-operator/blob/master/deploy/olm/storageos/storageos.package.yaml#L4 only.

deploy/olm-operators.configmap.yaml
```diff
@@ -292,11 +292,11 @@ data:
         minKubeVersion: "1.10.0"
         maturity: alpha
```
darkowlzz: Not to worry about changing this. I'm changing this file in another PR I'm working on.

deploy/olm/storageos/storageosoperator.0.0.10.clusterserviceversion.yaml
```diff
@@ -9,24 +9,93 @@ metadata:
...
   # replaces: 
   displayName: Storageos Cluster Operator
   description: |
-    StorageOS transforms commodity server or cloud based disk capacity into enterprise-class storage to run persistent workloads such as databases in containers. The StorageOS Operator installs and manages StorageOS clusters.
+    StorageOS is a software-defined storage platform that transforms commodity
+    server or cloud based disk capacity into enterprise-class storage for
+    running persistent workloads such as databases in containers.
+
+    The StorageOS Cluster Operator installs and manages StorageOS within an
+    Openshift cluster. Cluster nodes may contribute local or attached disk-based
```
actionbuddha: Is this Openshift only or all K8S derivatives?

croomes: Probably more targeted at Openshift at this point.

darkowlzz: All the development and tests, including e2e tests, were done on k8s.

PR #72, opened 2019-03-06
cluster-ctrl: reset cluster deployment with current cluster

darkowlzz:
This resets deployment along with the current cluster. Without this, the
deployment configuration was cached and reused if clusters with
different configuration are created consecutively.

This is needed to detect any future configuration caching bug.
It also reduces the number of parallel builds.


darkowlzz:
Reduces the cumulative build time by ~16 minutes :)

PR #78, opened 2019-03-11
Add csinodeinfos rbac to attacher

croomes:
The csi-attacher pod complains that the service-account it uses doesn't have access to csinodeinfos.

Before merging need to check whether it validates on k8s versions before CSINodeInfo was introduced (Alpha in 1.13)


darkowlzz:
Thanks. I've added the same permission for the operator so that it can grant those same permissions to the attacher.

darkowlzz:
> Before merging need to check whether it validates on k8s versions before CSINodeInfo was introduced (Alpha in 1.13)

It did work fine on k8s 1.11 as per the CI.

PR #79, opened 2019-03-17
cluster: fix cluster status update

darkowlzz:
Combine current cluster and deployment in to a single data type
StorageOSCluster so that the deployment always points to the right
cluster.

Also, adds status members checks in e2e test and enables go race detector
for the unit tests.


pkg/storageos/daemonset.go
```diff
@@ -0,0 +1,262 @@
...
+									},
+								},
+								{
+									Name:  joinEnvVar,
+									Value: s.stos.Spec.Join,
+									// ValueFrom: &corev1.EnvVarSource{
+									// 	FieldRef: &corev1.ObjectFieldSelector{
+									// 		FieldPath: "status.podIP",
+									// 	},
+									// },
```
croomes: Remove?

darkowlzz: Yes, we should remove that. We will not need that again.

PR #82, opened 2019-03-25
Update OLM CSV to v0.0.12

darkowlzz:


croomes:
When should we switch to version 1.0.0?

darkowlzz:
> When should we switch to version 1.0.0?

@croomes as soon as we upgrade the API version from `v1alpha1` to `v1`.

PR #86, opened 2019-03-25
Move API version to v1

darkowlzz:
This changes the API version of all the CRDs from v1alpha1 to v1.


deploy/storageos-operators.configmap.yaml
```diff
@@ -463,7 +463,7 @@ data:
 
     - packageName: storageos
       channels:
-      - name: alpha
+      - name: stable
         currentCSV: storageosoperator.0.0.0
```
croomes: Is `0.0.0` correct?

darkowlzz:
`deploy/examples.json` and `deploy/olm/storageos/` still have `v1alpha1` api version because they are related to the OLM CSV file which target a specific version of the operator. They should be updated once we have a v1 api version release available.

deploy/storageos-operators.configmap.yaml
darkowlzz: yes, this is used in the e2e test. We need not increment it every time.

actionbuddha:
Adding @avestuk  and @Arau for commentary and awareness - guys, be aware this is a breaking change.

domodwyer:
I noticed the `Default Value` for `images.nodeContainer` on the readme says `storageos/node:1.1.0` - I guess this should be `1.1.4` and changing it part of the release checklist @actionbuddha ?

PR #87, opened 2019-03-26
Export rbac resource names

darkowlzz:
Exporting the resource names makes is easy to refer the resources in
different packages.
Also renames roles and role bindings to the new format.

NOTE: Will export other resource names in follow up PRs. Since there's change in naming format for roles, doing it first.

### DIFF
--- a/pkg/storageos/rbac.go
+++ b/pkg/storageos/rbac.go
@@ -6,6 +6,28 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Exported role, binding and service account resource names.
+const (
+	DaemonsetSA   = "storageos-daemonset-sa"
+	StatefulsetSA = "storageos-statefulset-sa"
+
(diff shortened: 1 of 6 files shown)


pkg/storageos/rbac.go
```diff
@@ -6,6 +6,28 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Exported role, binding and service account resource names.
+const (
+	DaemonsetSA   = "storageos-daemonset-sa"
+	StatefulsetSA = "storageos-statefulset-sa"
+
+	CSIProvisionerClusterRoleName    = "storageos:csi-provisioner-role"
+	CSIProvisionerClusterBindingName = "storageos:csi-provisioner-binding"
```
croomes: I think we can remove the `-role` and `-binding` suffixes?

darkowlzz: Right, we can. I'll change it.

PR #91, opened 2019-04-01
Add Node Controller for copying node labels

darkowlzz:
Node Controller ensures that the k8s node labels are copied to storageos node labels whenever there's a change in any node.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -1,0 +1,223 @@
+package node
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	storageosv1 "github.com/storageos/cluster-operator/pkg/apis/storageos/v1"
+	storageosapi "github.com/storageos/go-api"
+	storageostypes "github.com/storageos/go-api/types"
+)
+
+var log = logf.Log.WithName("controller_node")
+
+// Add creates a new Node Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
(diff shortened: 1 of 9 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,251 @@
...
+	if len(node.Labels) == 0 {
+		node.Labels = make(map[string]string)
+	}
+
+	// Check if the k8s node labels already exist in storageos node labels.
+	// If there's no update, or addition, do not update the storageos node.
+	changed := false
+	for k, v := range labels {
+		// If the label already exists, compare the new and old values.
+		if v2, ok := node.Labels[k]; ok {
```
domodwyer: Just checking, as this is additive a label will never be removed right? 

I am a little worried about:

* Add label `replicas=3`
* Change your mind
* Remove label

I imagine we don't want to remove labels set that aren't StorageOS labels? Perhaps we can do some sort of set diff and remove entries in `node.Labels` that are not in `labels` that are prefixed by `storageos.com` only.

darkowlzz: Yes, we never remove labels.
AFAIK, we don't have any internally used labels for nodes. There are volume and pool labels https://docs.storageos.com/docs/reference/labels.
In order to avoid any stale node labels in the future, there had been discussions to restrict who can modify node labels, like allowing only the operator to modify node labels. But for now, we are only adding the labels and updating any existing label with new values.

croomes: We're double-checking what the delete behaviour should be.  The user-agent has been set, so we could decide to only allow the operator to modify node labels.

As it is currently, if you want to delete a node label that was inherited from k8s, you need to remove it from k8s and then remove it from the StorageOS node.  I think that's ok for initial version.

domodwyer: As briefly touched on in the CP meeting earlier it should be possible to do something like:

```go
// Global state in the operator
var (
	// prevNodeLabels maps node names to a set of previously observed kubernetes
	// labels for that node.
	//
	// It should only be accessed when holding the prevNodeLabelsMu lock (this
	// should probably be encapsulated into a type to only allow locked access).
	prevNodeLabels   = map[string]map[string]string{}
	prevNodeLabelsMu = &sync.Mutex{}
)

func syncLabels(name string, etc ...) {
	// Grab the old labels, and compare them to the new labels.
	//
	// This diff only operates on the k8s labels, so custom StorageOS labels
	// applied directly (not through k8s) remain unchanged.
	prevNodeLabelsMu.Lock()
	oldLabels := prevNodeLabels[name]
	prevNodeLabelsMu.Unlock()

	changes := diff(newLabels, oldLabels)

	// Apply changes
}
```

This lets the k8s labels be freely changed by the user of k8s, including removing them when they're removed in k8s, and it doesn't touch any labels set via StorageOS directly. 

Obviously it would have to be reset when the cluster changes, and it wouldn't work for cold-starts - in that case we would have to fall back to the additive-only state (or we could read the current labels at operator startup if we want to give better guarantees?).

All of this only works if the calls to `Reconcile` are linearised/single-threaded per node, and `oldLabels` is obviously a map that needs it's own lock if there can be concurrent calls to syncLabels for a given node - perhaps a struct that combines a lock and the map would be best as the value of the `prevNodeLabels` map.

Realise this might be not desired right now, but would be a nice usability addition to keep in mind!

croomes: Discussed with @darkowlzz - we think this will work well and we'd like to add, but in a separate PR.  Planning to do it next week.

PR #93, opened 2019-04-05
storageos: Add etcd mTLS support

darkowlzz:
This adds etcd mTLS support in the storageos deployment.

Also moves the constants and functions used in daemonset from deploy
file to daemonset file.


darkowlzz:
Rebased.

avestuk:
I've tested the changes and they work.

PR #95, opened 2019-04-08
[WIP] storageos: Add compute-only deployment support

darkowlzz:


croomes:
I'm not sure I like having separate DaemonSets for compute and storage nodes.  We should be able to pass selectors and determine dynamically?  This may be ok for now, but let's have a look at whether the current label implementation should be revised.

croomes:
Now that k8s node labels are passed to StorageOS, we should be able to set `storageos.com/deployment=computeonly` on the k8s nodes to enable compute only.  This lets us have a single DaemonSet with mixed node configs.  This wasn't clear in the docs before (fixed recently) https://docs.storageos.com/docs/reference/labels

PR #103, opened 2019-04-18
Operator release tools and docs

darkowlzz:
Release tools and instructions for operator community hub and rhel operator releases.


deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
```diff
@@ -34,7 +34,7 @@ metadata:
             "namespace": "default"
           },
           "spec": {
-            "image": "registry.connect.redhat.com/storageos/cluster-operator:1.0.0",
+            "image": "registry.connect.redhat.com/storageos/cluster-operator:latest",
```
croomes: Curious why it uses a mix of fixed version and latest?

darkowlzz: This is part of the example we provide. We don't want to be updating the example for every release.

PR #104, opened 2019-04-18
Metadata for cluster

croomes:
Cluster done, now the other CRDs.

### DIFF
--- a/deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
+++ b/deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
@@ -22,6 +22,7 @@ metadata:
             "namespace": "default"
           },
           "spec": {
+            "namespace": "kube-system",
             "secretRefName": "storageos-api",
             "secretRefNamespace": "default"
           }
@@ -49,7 +50,7 @@ metadata:
             "namespace": "default"
           },
           "spec": {
-            "newImage": "registry.connect.redhat.com/storageos/cluster-operator:latest"
+            "newImage": "registry.connect.redhat.com/storageos/node:latest"
           }
         }
       ]
@@ -124,8 +125,11 @@ spec:
 
     ## Required Parameters
 
-    * `secretRefName` - the name of a secret that contains two keys for the `apiUsername` and `apiPassword` to be used as api credentials ([documentation](https://docs.storageos.com/docs/reference/cluster-operator/examples))
-    * `secretRefNamespace` - the namespace where the api credentials secret is stored
+    * `secretRefName` - the name of a secret that contains two keys for the
+    `apiUsername` and `apiPassword` to be used as api credentials
+    ([documentation](https://docs.storageos.com/docs/reference/cluster-operator/examples))
+    * `secretRefNamespace` - the namespace where the api credentials secret is
+    stored
 
     ## Required SCC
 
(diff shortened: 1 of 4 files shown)


croomes:
also need to copy for csv-rhel

deploy/olm/storageos/storageos.clusterserviceversion.yaml
```diff
@@ -318,14 +321,90 @@ spec:
...
+      description: StorageOS Cluster installs StorageOS in the cluster. It
+        contains all the configuration for setting up a StorageOS cluster and
+        also shows the status of the running StorageOS cluster.
       specDescriptors:
+      - description: Defines the various container images used in the cluster.
+        displayName: Images
+        path: images        
+      - description: The namespace to install the StorageOS cluster into.
+          `kube-system` is recommended so that StorageOS does not get evicted if
+          a node becomes over-allocated.
```
darkowlzz: Should we mention what's the default value?

deploy/olm/storageos/storageos.clusterserviceversion.yaml
```diff
@@ -318,14 +321,90 @@ spec:
...
+      - description:
+        displayName:
+        path: disableFencing
+      - description: When Pod Fencing is disabled, StorageOS will not perform
+          any interaction with Kubernetes when it detects that a node has gone
+          offline. Additionally, the Kubernetes permissions required for Fencing
+          will not be added to the StorageOS role.
+        displayName: Disable Fencing
+        path: disableFencing
+      - description: When enabled, the Operator will not perform any actions.
```
darkowlzz: Will not perform any action on the cluster. It's a cluster specific thingy. But since we allow only a single cluster to run at a time, it's the main cluster.

deploy/olm/storageos/storageos.clusterserviceversion.yaml
croomes: yes, thought the default is still `storageos`.  Should I change to `kube-system`?

darkowlzz: I'm not sure if just moving to `kube-system` will be enough. We need to set pod priority class to ensure we don't get evicted under resource pressure.

croomes: I think for now, we should keep the wording, then in a separate PR we change the default and add the priority class?

deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
```diff
@@ -347,9 +427,51 @@ spec:
...
+        StorageOS cluster as per the upgrade configuration.
+      specDescriptors:
+      - description: The StorageOS Node image to upgrade to. 
+          e.g. `registry.connect.redhat.com/storageos/node:latest`
+        displayName: New Image
+        path: newImage
+      statusDescriptors:
+      - description: Set to true if the upgrade completed.
+        displayName: Completed
+        path: completed
```
darkowlzz: No newline at the end of the file.

deploy/olm/storageos/storageos.clusterserviceversion.yaml
```diff
@@ -347,9 +427,51 @@ spec:
...
+        StorageOS cluster as per the upgrade configuration.
+      specDescriptors:
+      - description: The StorageOS Node image to upgrade to. 
+          e.g. `storageos/node:latest`
+        displayName: New Image
+        path: newImage
+      statusDescriptors:
+      - description: Set to true if the upgrade completed.
+        displayName: Completed
+        path: completed
```
darkowlzz: newline at the end.

deploy/olm/storageos/storageos.clusterserviceversion.yaml
croomes: Added to the example instead of changing the default, which would be a breaking change.

PR #109, opened 2019-04-21
Automate OLM based releases

darkowlzz:
Automates OLM based releases for operator to avoid error-prone manual artifact creation.
- Publish OLM metadata as github releases - The metadata zip is posted on the github releases page.
- Trigger rhel container service build - Build, scan and release new image
- Create PR to update community operator version

With this, a new tag must only be created once all the CRD, CSV and package files are updated for the new version. No two releases can have same CSV file. Every release must have a new CSV file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ jobs:
             tags: true
         - provider: releases
           api_key:
-            secure: # Add encrypted Github API token with repo permission here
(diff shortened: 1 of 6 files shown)


darkowlzz:
Added a release process doc at docs/release-process.md and added the required keys in travisCI configurations. This is ready for use now.

croomes:
Thanks for adding the extra docs, I think we'll be thankful for them later :)

PR #112, opened 2019-04-26
Add k8sDistro parameter and pass to DaemonSet

croomes:
Allows the Kubernetes distribution name to be set during provisioning, and passes it to the node container using the `K8S_DISTRO` env var.

This will be used for pre-configuring for specific distributions and for telemetry.

### DIFF
--- a/deploy/olm/storageos/storageos.clusterserviceversion.yaml
+++ b/deploy/olm/storageos/storageos.clusterserviceversion.yaml
@@ -386,6 +386,11 @@ spec:
           StorageOS pods.
         displayName: Tolerations
         path: tolerations
+      - description: Name of the Kubernetes distribution in use, e.g.
+          `openshift`.  This will be included in the product telemetry (if
+          enabled), to help focus development efforts.
+        displayName: Kubernetes Distribution Name
+        path: k8sDistro
       - description: To disable anonymous usage reporting across the cluster,
           set to true. Defaults to false. To help improve the product, data such
           as API usage and StorageOS configuration information is collected.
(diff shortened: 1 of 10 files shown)


pkg/storageos/daemonset.go
```diff
@@ -37,6 +37,7 @@ const (
 	kvAddrEnvVar                        = "KV_ADDR"
 	kvBackendEnvVar                     = "KV_BACKEND"
 	debugEnvVar                         = "LOG_LEVEL"
+	k8sDistroEnvVar                     = "k8S_DISTRO"
```
darkowlzz: `K8S_DISTRO`

deploy/olm/storageos/storageos.clusterserviceversion.yaml
```diff
@@ -386,6 +386,11 @@ spec:
           StorageOS pods.
         displayName: Tolerations
         path: tolerations
+      - description: Name of the Kubernetes distribution in use, e.g.
+          `openshift`.  This will be included in the product telemetry (if
+          enabled), to help focus development efforts.
+        displayName: Kubernetes Distribution Name
+        path: k8sDistro
```
darkowlzz: Should also add in `/deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml`

pkg/storageos/daemonset.go
croomes: well spotted, thanks!

deploy/olm/storageos/storageos.clusterserviceversion.yaml
croomes: doh! missed one...

darkowlzz:
rebased

PR #115, opened 2019-04-30
test: update kernel module name

darkowlzz:
TravisCI switched to use ubuntu xenial as default. The previous kernel
module package name is no longer available. Update the kernel module
package name.


darkowlzz:
Looks like it's still rolling out.
Builds for this project is still at `Runtime kernel version: 4.4.0-101-generic`.

darkowlzz:
It's time for this https://travis-ci.org/storageos/cluster-operator/jobs/536286792#L12
TravisCI has updated our builds to use gcp kernel.

PR #117, opened 2019-05-01
Add OCP SCC support

darkowlzz:
Grant OCP(OpenShift Container Platform) SCC permission to the operator at deployment and use the
permission in operator to grant SCC permission to the daemonset and
statefulset.

The OCP specific rules are ignore when used in upstream k8s.

### DIFF
--- a/pkg/storageos/delete.go
+++ b/pkg/storageos/delete.go
@@ -1,5 +1,7 @@
 package storageos
 
+import "strings"
+
 // Delete deletes all the storageos resources.
 // This explicit delete is implemented instead of depending on the garbage
 // collector because sometimes the garbage collector deletes the resources
@@ -74,6 +76,17 @@ func (s *Deployment) Delete() error {
 		}
 
 		if err := s.deleteCSISecrets(); err != nil {
+			return err
+		}
+	}
+
+	// Delete cluster role for openshift security context constraints.
+	if strings.Contains(s.stos.Spec.K8sDistro, k8sDistroOpenShift) {
+		if err := s.deleteClusterRoleBinding(OpenShiftSCCClusterBindingName); err != nil {
+			return err
+		}
+
+		if err := s.deleteClusterRole(OpenShiftSCCClusterRoleName); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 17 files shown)


pkg/storageos/delete.go
```diff
@@ -78,6 +80,17 @@ func (s *Deployment) Delete() error {
 		}
 	}
 
+	// Delete cluster role for openshift security context constraints.
+	if strings.Contains(s.stos.Spec.K8sDistro, k8sDistroOpenShift) {
```
croomes: I like using the k8sDistro here, but how does it get set?

darkowlzz: We have to set it in the cluster config while deploying.
We should change the CR example for cluster to have k8sDistro set to `openshift` for rhel operator release.

darkowlzz:
Added OLM e2e test on openshift and removed the execution of manual SCC commands from the openshift setup.

Install [yq](http://mikefarah.github.io/yq/) for yaml processing. This can be used to automate some of the configuration files based on the CI runs. For the openshift environment, the cluster CR is modified to have `k8sDistro: openshift`.

The operator-sdk e2e tests have `k8sDistro` set to `openshift` even if they are run on upstream k8s, still works fine. This works fine on openshift when run locally but fails in the CI. Although the OLM tests, using the same SCC permission grant works properly. Need to try to reproduce it again and investigate the issue.

darkowlzz:
Unable to reproduce the openshift operator-sdk e2e test failure. Keeping the SCC commands at the very end when the failing test is run for now. Maybe we will not need this test once we remove statefulset deployment. But the same statefulset resources are being tested in the OLM tests. Any issues should be caught there.

darkowlzz:
Switching the test order and running CSI test before native driver test fixed the openshift failures. The SCC rule with daemonset permission only in native driver test wasn't being updated when CSI test ran after it. Hence the failure was partial, daemonset was ready but not the statefulset. Changing the order and creating SCC rule with both daemonset and statefulset first fixed the issue.
Need to investigate ways to cleanup the SCC rules after every test run using `oc` cli tool.

PR #120, opened 2019-05-03
Automate OLM metadata update and verify

darkowlzz:
Uses yq(yaml processor) to automate metadata file generation and verification.
Any change in OLM should be made in the OLM configmap and run `make metadata-update` to update all the associated files automatically.
CI runs `make olm-lint` which generates the latest metadata files and compares them with the checked-in files. Any differences in the metadata files will result in a failure.

None of the files in `deploy/olm/{csv-rhel,storageos}/` should be touched manually.
Distribution/release specific updates must be added to `deploy/olm/{rhel-changes,community-changes,package-changes}.yaml`. Running the update script will update all the files accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,15 @@ metadata-zip:
 		deploy/olm/storageos/storageosupgrade.crd.yaml \
 		deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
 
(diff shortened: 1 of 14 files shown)


darkowlzz:
Need to add a README for the whole process.

PR #127, opened 2019-05-17
e2e: Update KinD to latest with k8s 1.14.2

darkowlzz:
Updates KinD to the latest version(1d85e92) which uses containerd
directly instead of docker. This improves the startup time of k8s
cluster.
The new k8s 1.14.2 container image compatible with KinD is available
at storageos/kind-node:v1.14.2. This image is built with /sys as shared
mount, which is required for storageos to run.

Also add KinD node image building docs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ jobs:
       env:
         - "TEST_CLUSTER=kind"
         - "INSTALL_METHOD=olm"
-      name: OLM on KinD (k8s-1.14)
+      name: OLM on KinD (k8s-1.13)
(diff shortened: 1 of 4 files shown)


darkowlzz:
OLM framework installation fails on the new KinD setup with containerd on k8s 1.14.2, on TravisCI environment only. Works fine when run locally. The olm-operators pod fails health check https://travis-ci.org/storageos/cluster-operator/jobs/533739377#L2079

For now, conditionally running the old version of KinD with k8s 1.13.2 for kind-olm test only. All other KinD based test setup are using the new KinD with containerd and k8s 1.14.2.

PR #131, opened 2019-05-27
Add operator quick install manifest

darkowlzz:
This adds `create-manifest.sh` to generate a single manifest file for
installing the operator. This will be used by the CI to generate a quick
install manifest file for every release.

With this, it'll be possible to install the operator by doing:
```
$ kubectl create -f https://github.com/storageos/cluster-operator/releases/download/<version>/storageos-operator.yaml
customresourcedefinition.apiextensions.k8s.io/storageosclusters.storageos.com created
customresourcedefinition.apiextensions.k8s.io/storageosupgrades.storageos.com created
customresourcedefinition.apiextensions.k8s.io/jobs.storageos.com created
namespace/storageos-operator created
clusterrole.rbac.authorization.k8s.io/storageos-operator created
serviceaccount/storageoscluster-operator-sa created
clusterrolebinding.rbac.authorization.k8s.io/storageoscluster-operator-rolebinding created
deployment.apps/storageos-cluster-operator created
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,10 @@
 build/_output
 build/_test
 build/operator-sdk
+build/yq
 deploy/test
 test/olm
+storageos-operator.yaml
 # Created by https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 ### Emacs ###
 # -*- mode: gitignore; -*-
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,7 @@ jobs:
       script:
         - make image/cluster-operator
         - make metadata-zip
+        - OPERATOR_IMAGE=storageos/cluster-operator:$TRAVIS_TAG make generate-install-manifest
       before_deploy:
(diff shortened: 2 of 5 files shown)


croomes:
This is a really good idea and worked well.  A few small things (none blockers):
1. check for `yq` and link to install page (https://mikefarah.github.io/yq/) if not found
2. create `storageos-operator` namespace if it doesn't exist
3. `make build/cluster-operator` supports setting image name with `OPERATOR_IMAGE` env var, but `make generate-install-manifest` takes an argument to set `OPERATOR_TAG`.  Can we use the `OPERATOR_IMAGE` method for both?  For testing it's useful not to use `storageos/cluster-operator` and use a private registry account instead.

darkowlzz:
Added a make target to install yq when generating the manifest and also moved the manifest to project root. Also, changed the env var to `OPERATOR_IMAGE`, like the other build target.

PR #135, opened 2019-06-21
Add storageos scheduler deployment

darkowlzz:
This adds deployment of an external scheduler using kube-scheduler as a
Deployment. This is automatically deployed by default. For a workload to
use storageos scheduler, the pod spec must have `schedulerName` set to
`storageos-scheduler`.

Adds a new StorageOSCluster spec property `disableScheduler` to disable
the scheduler. Also adds `image.hyperkubeContainer` to set the hyperkube
container image to use. By default, `gcr.io/google_containers/hyperkube`
is the container image used for hyperkube. The image tag depends on
the version of kubernetes cluster running.

### DIFF
--- a/pkg/apis/storageos/v1/storageoscluster_types.go
+++ b/pkg/apis/storageos/v1/storageoscluster_types.go
@@ -39,6 +39,8 @@ const (
 	CSIv0ExternalProvisionerContainerImage    = "storageos/csi-provisioner:v0.4.2"
 	CSIv0ExternalAttacherContainerImage       = "quay.io/k8scsi/csi-attacher:v0.4.2"
 
+	DefaultHyperkubeContainerRegistry = "gcr.io/google_containers/hyperkube"
+
 	DefaultPluginRegistrationPath = "/var/lib/kubelet/plugins_registry"
 	OldPluginRegistrationPath     = "/var/lib/kubelet/plugins"
 
@@ -185,6 +187,9 @@ type StorageOSClusterSpec struct {
 	// distribution information will also be included in the product telemetry
 	// (if enabled), to help focus development efforts.
 	K8sDistro string `json:"k8sDistro"`
+
+	// Disable StorageOS scheduler extender.
+	DisableScheduler bool `json:"disableScheduler"`
 }
 
 // StorageOSClusterStatus defines the observed state of StorageOSCluster
@@ -305,6 +310,17 @@ func (s StorageOSClusterSpec) GetCSILivenessProbeImage() string {
 		return s.Images.CSILivenessProbeContainer
 	}
 	return CSIv1LivenessProbeContainerImage
+}
+
+// GetHyperkubeImage returns hyperkube container image for a given k8s version.
+// If an image is set explicitly in the cluster configuration, that image is
+// returned.
+func (s StorageOSClusterSpec) GetHyperkubeImage(k8sVersion string) string {
+	if s.Images.HyperkubeContainer != "" {
+		return s.Images.HyperkubeContainer
+	}
+	// Add version prefix "v" in the tag.
+	return fmt.Sprintf("%s:v%s", DefaultHyperkubeContainerRegistry, k8sVersion)
 }
 
 // GetServiceName returns the service name.
@@ -460,6 +476,7 @@ type ContainerImages struct {
 	CSIExternalProvisionerContainer    string `json:"csiExternalProvisionerContainer"`
 	CSIExternalAttacherContainer       string `json:"csiExternalAttacherContainer"`
 	CSILivenessProbeContainer          string `json:"csiLivenessProbeContainer"`
+	HyperkubeContainer                 string `json:"hyperkubeContainer"`
 }
 
 // StorageOSClusterCSI contains CSI configurations.
--- a/pkg/storageos/csi_helper.go
+++ b/pkg/storageos/csi_helper.go
@@ -248,49 +248,15 @@ func (s Deployment) csiHelperVolumes() []corev1.Volume {
 	}
 }
 
-// getCSIHelperStatefulSet returns the CSI helper StatefulSet resource.
-func (s Deployment) getCSIHelperStatefulSet(name string) *appsv1.StatefulSet {
-	return &appsv1.StatefulSet{
-		TypeMeta: metav1.TypeMeta{
-			APIVersion: "apps/v1",
-			Kind:       "StatefulSet",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
-			Namespace: s.stos.Spec.GetResourceNS(),
-			Labels: map[string]string{
-				"app": "storageos",
-			},
-		},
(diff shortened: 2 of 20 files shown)


pkg/storageos/scheduler_extender.go
```diff
@@ -0,0 +1,280 @@
...
+	// Create RBAC related resources.
+	if err := s.createClusterRoleForScheduler(); err != nil {
+		return err
+	}
+	if err := s.createServiceAccountForScheduler(); err != nil {
+		return err
+	}
+	if err := s.createClusterRoleBindingForScheduler(); err != nil {
+		return err
+	}
```
darkowlzz: Creating all these associated components here to reduce cluttering `Deploy()`.
I'll change CSI helper deployment similar to this in a separate PR.

pkg/storageos/scheduler_extender.go
```diff
@@ -0,0 +1,280 @@
...
+	if err := s.deleteClusterRoleBinding(SchedulerClusterBindingName); err != nil {
+		return err
+	}
+	if err := s.deleteServiceAccount(SchedulerSA); err != nil {
+		return err
+	}
+	if err := s.deleteClusterRole(SchedulerClusterRoleName); err != nil {
+		return err
+	}
+	return nil
```
darkowlzz: Deleting all the related resources here to avoid cluttering `Delete()`.
Will change the same for CSI helpers in a separate PR.

pkg/storageos/deploy.go
```diff
@@ -302,6 +310,40 @@ func (s *Deployment) deleteObject(obj runtime.Object) error {
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: s.stos.Spec.GetResourceNS(),
+			Labels: map[string]string{
+				"app": "storageos",
+			},
+		},
+	}
+}
```
darkowlzz: Moved and renamed these functions to be more generic than just CSI helper specific.

pkg/storageos/csi_helper.go
```diff
@@ -248,49 +248,15 @@ func (s Deployment) csiHelperVolumes() []corev1.Volume {
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
-			Namespace: s.stos.Spec.GetResourceNS(),
-			Labels: map[string]string{
-				"app": "storageos",
-			},
-		},
-	}
-}
-
```
darkowlzz: Moved to deploy.go and renamed to not be CSI helper specific.

pkg/storageos/scheduler_extender.go
```diff
@@ -0,0 +1,280 @@
...
+
+// schedulerVolumes returns a list of volumes that should be part of the
+// scheduler extender deployment.
+func (s Deployment) schedulerVolumes() []corev1.Volume {
+	return []corev1.Volume{
+		{
+			Name: "storageos-scheduler-config",
+			VolumeSource: corev1.VolumeSource{
+				ConfigMap: &corev1.ConfigMapVolumeSource{
+					LocalObjectReference: corev1.LocalObjectReference{Name: "storageos-scheduler-config"},
```
darkowlzz: Should use `schedulerConfigConfigMapName` here.

pkg/storageos/scheduler_extender.go
```diff
@@ -0,0 +1,280 @@
...
+      leaderElect: {{.LeaderElection}}
+      lockObjectName: {{.SchedulerName}}
+      lockObjectNamespace: {{.Namespace}}
+`
+	schedConfigData := schedulerConfigTemplate{
+		SchedulerName:  schedulerExtenderName,
+		PolicyName:     policyConfigMapName,
+		Namespace:      s.stos.Spec.GetResourceNS(),
+		LeaderElection: true,
+	}
```
croomes: Curious how the leader election works?  I saw you've set replicas to 1 - do we need to test > 1?

darkowlzz: I think we should provide an option in the cluster config to set this number, defaulting to 1 or 2. I've not looked into leader election. We do need to test it. Also, maybe we should add pod toleration for quick recovery, similar to the ones we have for CSI helper deployment pods.

pkg/apis/storageos/v1/storageoscluster_types.go
```diff
@@ -291,6 +296,17 @@ func (s StorageOSClusterSpec) GetCSILivenessProbeImage() string {
...
 
+// GetHyperkubeImage returns hyperkube container image for a given k8s version.
+// If an image is set explicitly in the cluster configuration, that image is
+// returned.
+func (s StorageOSClusterSpec) GetHyperkubeImage(k8sVersion string) string {
+	if s.Images.HyperkubeContainer != "" {
+		return s.Images.HyperkubeContainer
+	}
+	// Add version prefix "v" in the tag.
+	return fmt.Sprintf("%s:v%s", DefaultHyperkubeContainerRegistry, k8sVersion)
```
Arau: I found that for many clusters the image tag :k8sVersion results in a pull failure as the image doesn't exist.

For instance the following cluster definition tries to pull `gcr.io/google_containers/hyperkube:v1.12.6-eks-d69f1b` as the `k8sVersion` is `1.12.6-eks-d69f1b`. Image that doesn't exist. 

```
serverVersion:
  buildDate: 2019-02-28T20:26:10Z
  compiler: gc
  gitCommit: d69f1bf3669bf00b7f4a758e978e0e7a1e3a68f7
  gitTreeState: clean
  gitVersion: v1.12.6-eks-d69f1b
  goVersion: go1.10.8
  major: "1"
  minor: 12+
  platform: linux/amd64
```

darkowlzz: ah! So those are commit hashes. We can trim the version and extract the first tri-dotted part of the version.
I'll add something to handle this with tests.
If you find more variants, please share about them.

Arau: This issue can be mitigated with documentation by defining an explicit image in the CR for specific providers in case that you consider that the `k8sVersion` variable should follow the standard.

Arau: I found that Rancher clusters deployed with a Driver also use the hash of the commit. Without using a driver, the k8sVersion is as expected
```
serverVersion:
  buildDate: 2019-06-19T20:50:07Z
  compiler: gc
  gitCommit: fcbc1d20b6bca1936c0317743055ac75aef608ce
  gitTreeState: clean
  gitVersion: v1.13.6-gke.13
  goVersion: go1.11.5b4
  major: "1"
  minor: 13+
  platform: linux/amd64
```

Arau: I don't know if its feasible but the `major` + `minor` seems more stable for what we are trying to do.

croomes:
Tested and couldn't find any issues

PR #137, opened 2019-06-25
Fix metadata zip generator

darkowlzz:
Operator metadata zip that consists of CSVs with replaces field must
also include the version of CSV that's being replaced.
This change packages the versioned CSV files in the zip.

A reference to default old and current versions of OLM releases are
added in the Makefile, which is updated by the release-gen.sh script
when preparing for a new release.


darkowlzz:
I've uploaded a new zip file, generated using these changes, at https://github.com/storageos/cluster-operator/releases/tag/1.2.0.

PR #146, opened 2019-07-25
storageos: Add scheduler perm for events patch

darkowlzz:
Scheduler shows the following error because it can't patch events:
`forbidden: User
"system:serviceaccount:storageos:storageos-scheduler-sa" cannot
patch resource "events" in API group "" in the namespace... `.
This change add patch permission.


darkowlzz:
CI caught the issues with the changes. Needed some more permissions and also generate the olm files with new permissions.

PR #148, opened 2019-07-29
Generate openapi based CRD spec

darkowlzz:
Use operator-sdk generate openapi and kubebuilder tags to generate the
CRD spec from go definitions of the custom resources.

Also, generate the spec before linting to ensure olm metadata are
updated whenever there's a change in CRD spec.


darkowlzz:
Have to disable CI for k8s 1.10 because CRD validation in 1.10 does not fully supports openAPI schema. k/k#59485.

PR #151, opened 2019-08-09
Add support for NFS server

darkowlzz:
Adds a new custom resource `nfsservers.storageos.com` for managing NFS server resources.

A StorageOS Cluster resource is a prerequisite for NFS server because the underlying storage requires StorageOS volumes.
On creating a NFSServer resource, the NFS server controller provisions NFS ganesha server instance with a StorageOS volume attached.

Short name for `nfsservers.storageos.com` is `nfsserver`.

Status of provisioned NFS server:
```
$ kubectl get nfsserver
NAME                STATUS    CAPACITY   TARGET         ACCESS MODES   STORAGECLASS   AGE
example-nfsserver   Running   1Gi        10.103.56.29   RW             fast           52m
```


pkg/nfs/configmap.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, export := range instance.Spec.Exports {
+		exportCfg, err := exportConfig(id, export.PersistentVolumeClaim.ClaimName, export.Server.AccessMode, export.Server.Squash)
+		if err != nil {
+			return "", err
+		}
+		exports = append(exports, exportCfg)
+		id++
+	}
+
+	globalCfg, err := globalConfig(true, true)
```
croomes: Use consts for graceless and fsidDevice?

pkg/nfs/configmap.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// Otherwise use export list
+	for _, export := range instance.Spec.Exports {
+		exportCfg, err := exportConfig(id, export.PersistentVolumeClaim.ClaimName, export.Server.AccessMode, export.Server.Squash)
+		if err != nil {
+			return "", err
+		}
+		exports = append(exports, exportCfg)
+		id++
+	}
```
croomes: I think we can simplify this a bit as we've settled on having a single export per pv.  I don't think we will need to support multiple exports per server.

pkg/nfs/deploy.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+const (
+	appName         = "storageos"
+	statefulsetKind = "statefulset"
+
+	DefaultNFSPort     = 2049
+	DefaultRPCPort     = 111
```
croomes: we can remove `DefaultRPCPort`.  Only 2049 and 9587 are needed.

pkg/nfs/statefulset.go
```diff
@@ -0,0 +1,167 @@
...
+					},
+					Ports: []corev1.ContainerPort{
+						{
+							Name:          "nfs-port",
+							ContainerPort: int32(nfsPort),
+						},
+						{
+							Name:          "rpc-port",
+							ContainerPort: int32(rpcPort),
+						},
```
croomes: can remove rpc-port

pkg/nfs/statefulset.go
```diff
@@ -0,0 +1,167 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: labelsForStatefulSet(d.nfsServer.Name, labels),
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					ImagePullPolicy: "IfNotPresent",
+					Name:            "nfsd",
+					Image:           d.nfsServer.Spec.GetContainerImage(),
+					// Args: []string{"nfs", "server", "--ganeshaConfigPath=" + NFSConfigMapPath + "/" + nfsServer.name},
```
croomes: Args can be removes - new container will only use env vars

pkg/nfs/statefulset.go
```diff
@@ -0,0 +1,167 @@
...
+					Env: []corev1.EnvVar{
+						{
+							Name:  "GANESHA_CONFIGFILE",
+							Value: "/config/" + d.nfsServer.Name,
+						},
+					},
```
croomes: Can you also add env vars for `NAME` and `NAMESPACE` set to the RWX PVC name and namespace?  This will be used for decorating the metrics.

pkg/apis/storageos/v1/nfsserver_types.go
```diff
@@ -0,0 +1,208 @@
...
+	// Resources represents the minimum resources required
+	// +optional
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// The annotations-related configuration to add/set on each Pod related object.
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Replicas of the NFS daemon
+	// TODO(simon): Don't think we can have multiple servers?
+	// Replicas int `json:"replicas,omitempty"`
```
croomes: can remove Replicas

pkg/apis/storageos/v1/nfsserver_types.go
```diff
@@ -0,0 +1,208 @@
...
+	// Reading and Writing permissions on the export
+	// Valid values are "ReadOnly", "ReadWrite" and "none"
+	AccessMode string `json:"accessMode,omitempty"`
+
+	// This prevents the root users connected remotely from having root privileges
+	// Valid values are "none", "rootid", "root", and "all"
+	Squash string `json:"squash,omitempty"`
+
+	// The clients allowed to access the NFS export
+	AllowedClients []AllowedClientsSpec `json:"allowedClients,omitempty"`
```
croomes: We should either remove `AllowedClients` or figure out how it might be used and test it.  I think it's probably best to remove it for now.

pkg/apis/storageos/v1/nfsserver_types.go
```diff
@@ -0,0 +1,208 @@
...
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// Constants for NFSServer default values and different phases.
+const (
+	// DefaultNFSContainerImage is the name of the Ganesha container to run.
+	// TODO: change to an image we maintain.
+	DefaultNFSContainerImage = "darkowlzz/nfs-ganesha:v0.0.2"
```
croomes: `storageos/nfs:test` is now pushed for testing.

darkowlzz:
CI was failing due to pod toleration.
```
  Warning  FailedScheduling  15s (x4 over 19s)  default-scheduler  0/1 nodes are available: 1 node(s) had taints that the pod didn't tolerate.
```
We taint the only node in order to test pod toleration.
Need to add pod toleration configuration for NFS server as well.

pkg/nfs/statefulset.go
darkowlzz: Exact PVC name won't be accurate. Because the volume claim is a template in StatefulSet, only the PVC name prefix is predictable. PVC name is appended by the StatefulSet name and a number for the replicas.
I guess the prefix name should be enough?

pkg/nfs/configmap.go
darkowlzz: Will remove this in a separate PR.

pkg/apis/storageos/v1/nfsserver_types.go
darkowlzz: Keeping it for now. Will remove in a follow-up PR.

darkowlzz:
Added a temporary check in e2e tests as a workaround for the limited CSI support in openshift 3.11, which is required for NFS server. CSI volumes mount fail and the NFS server pod remains in pending state. SCC permissions are added. Should work on OpenShift 4.
Addressed the review suggestions.

pkg/nfs/statefulset.go
croomes: It should be the name of the RWX vol, not the underlying StorageOS vol, so I think you can use `d.nfsServer.Name` and `d.nfsServer.Namespace` here.

pkg/nfs/configmap.go
darkowlzz: Remove multiple exports support and changed it to customizable single export in this PR itself.

pkg/nfs/statefulset.go
darkowlzz: Added the env vars in this PR itself.

pkg/apis/storageos/v1/nfsserver_types.go
darkowlzz: Removed it in this PR.

PR #157, opened 2019-09-04
Bump kind k8s version to 1.15.3

croomes:
storageos/kind-node:v1.15.3 adds the nfs-common package needed for NFS testing.  Not required for the operator.  Built from https://github.com/storageos/kind/tree/kind-23-08-19

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ go:
 env:
   global:
     - CHANGE_MINIKUBE_NONE_USER=true
-    - K8S_VERSION="v1.10.0"
+    - K8S_VERSION="v1.15.3"
     - MINIKUBE_VERSION="v0.28.2"
     - IMAGE_NAME=storageos/cluster-operator
     - IMAGE_TAG=test


.travis.yml
```diff
@@ -8,7 +8,7 @@ go:
 env:
   global:
     - CHANGE_MINIKUBE_NONE_USER=true
-    - K8S_VERSION="v1.10.0"
+    - K8S_VERSION="v1.15.3"
```
darkowlzz: This env var was used by minikube setup only.
For KinD, k8s version 1.13 and 1.14 are selectively used in https://github.com/storageos/cluster-operator/blob/1e252955535532638141ce65fb61321d5692aa6b/test/e2e.sh#L234-L236
1.13 was required for OLM only because OLM installation resulted in failing pods when deployed in 1.14 and I couldn't figure out what was causing that. So OLM tests used k8s 1.13 and operator go e2e test used 1.14. 1.14 can be replaced with 1.15.x. We can also try running OLM tests in 1.15.x and see if they work now. Only the OLM framework installation used to fail. Not storageos operator.

PR #158, opened 2019-09-04
OLM: update example with NFS, make CSI default

darkowlzz:


deploy/storageos-operators.configmap.yaml
```diff
@@ -1122,7 +1126,10 @@ data:
             version: v1
             kind: NFSServer
             displayName: NFS Server
-            description: NFS Server creates a NFS Ganesha server.
+            description: StorageOS NFS Server provides support for ReadWriteMany
```
croomes: In this line, please change "ReadWriteMany" -> "shared"

Issue #162, opened 2019-09-07
storageos operator missing pods/log permission


esara:
$ kubectl logs -f storageos-cluster-operator-598bbc4d79-k66qj  -n storageos-operator
2019-09-07T02:46:03.279Z	INFO	storageos.job	Creating a new DaemonSet	{"Request.Namespace": "default", "Request.Name": "storageos-job"}
2019-09-07T02:46:03.342Z	DEBUG	kubebuilder.controller	Successfully Reconciled	{"controller": "job-controller", "request": "default/storageos-job"}
2019-09-07T02:46:03.478Z	INFO	storageos.job	No DaemonSets found
2019-09-07T02:46:03.889Z	INFO	storageos.job	Failed to get logs from pod	{"pod": "storageos-job-daemonset-job-7876t", "error": "pods \"storageos-job-daemonset-job-7876t\" is forbidden: User \"system:serviceaccount:storageos-operator:storageoscluster-operator-sa\" cannot get resource \"pods/log\" in API group \"\" in the namespace \"default\""}
2019-09-07T02:46:03.891Z	INFO	storageos.job	Failed to get logs from pod	{"pod": "storageos-job-daemonset-job-lswzv", "error": "pods \"storageos-job-daemonset-job-lswzv\" is forbidden: User \"system:serviceaccount:storageos-operator:storageoscluster-operator-sa\" cannot get resource \"pods/log\" in API group \"\" in the namespace \"default\""}
2019-09-07T02:46:03.892Z	INFO	storageos.job	Failed to get logs from pod	{"pod": "storageos-job-daemonset-job-p4wvl", "error": "pods \"storageos-job-daemonset-job-p4wvl\" is forbidden: User \"system:serviceaccount:storageos-operator:storageoscluster-operator-sa\" cannot get resource \"pods/log\" in API group \"\" in the namespace \"default\""}

PR #163, opened 2019-09-07
fix for issue #162 - adding pods/log permission to the cluster-operator

esara:
cluster-operator can check the logs after the creation of the job


darkowlzz:
Hi, thanks for submitting a fix for this issue. `job` and `storageosupgrade` CRDs are not maintained for now. When they were added in the operator, operator used to have cluster admin role. Later specific permissions were added and cluster admin requirement was removed. Since `job` and `storageosupgrade` CRDs weren't maintained or tested, the permissions required for them have not been added in the operator. We will be adding automated tests and maintain those two CRDs in the coming months.
The fix looks good and it'll work when the operator is deployed using the manifest files. But we also support other means of installing the operator, like using OLM ([Operator Lifecycle Manager](https://github.com/operator-framework/operator-lifecycle-manager)). For that, can you please add the same change in [storageos-operators.configmap.yaml](https://github.com/storageos/cluster-operator/blob/17e8591165561a99627f1348c6a551fff5750100/deploy/storageos-operators.configmap.yaml#L849) and then run `make metadata-update` to populate the same change in all the various places it's required?
Also, would be great if you could squash all the commits into a single commit.

Thanks for reporting and working on it.

PR #168, opened 2019-09-18
Set NFS volume replica for pod fencing

darkowlzz:
In order for pod fencing to work, there must be at least one healthy
volume replica.


darkowlzz:
Adding a default replica results in volume provision failure when there's not enough nodes for replica.
It'll be better to keep this opt-in. Users can specify a StorageClass with replica for fencing and high availability needs.

croomes:
Yes, we discussed and agreed that we shouldn't set replicas by default.  Instead this can be set in the StorageClass or PVC labels.

PR #171, opened 2019-09-24
Upgrade to operator-sdk v0.10.0

darkowlzz:
Updates the k8s related dependencies to the versions pinned by operator-sdk v0.10.0.


Gopkg.toml
```diff
@@ -60,6 +60,10 @@ required = [
   name = "github.com/blang/semver"
   version = "v3.6.1"
 
+[[override]]
+  name = "github.com/coreos/prometheus-operator"
+  version = "v0.29.0"
```
darkowlzz: This is not needed here, but is included here for the future work that uses operator-sdk metrics.

darkowlzz:
This increases the scorecard test time due to the known reasons and possible solutions described in
https://github.com/storageos/cluster-operator/blob/master/test/scorecard-test.sh#L23-L28 . It can be addressed separately because the fix is not obvious and require refactoring of the controllers.

PR #172, opened 2019-09-26
NFS: Allow using existing PVC with NFS server

darkowlzz:
Adds a new field at the root of NFSServer spec `persistentVolumeClaim`
to specify a reference to an existing PVC.
Also, consider this specified PVC in the NFS export config generation.

Example of a NFSServer CR using an existing PVC:
```
apiVersion: storageos.com/v1
kind: NFSServer
metadata:
  name: example-nfsserver
spec:
  persistentVolumeClaim:
    claimName: pv1
    readOnly: false
```

Adds NFSServer CRD diff checker in the metadata-diff-checker.


pkg/nfs/statefulset.go
```diff
@@ -147,4 +148,18 @@ func (d *Deployment) createPodTemplateSpec(nfsPort int, httpPort int, labels map
...
+	// the volumes list.
+	if d.nfsServer.Spec.PersistentVolumeClaim.ClaimName != "" {
+		pvc := corev1.Volume{
+			Name: PVCNamePrefix,
+			VolumeSource: corev1.VolumeSource{
+				PersistentVolumeClaim: &d.nfsServer.Spec.PersistentVolumeClaim,
+			},
+		}
+		podSpec.Spec.Volumes = append(podSpec.Spec.Volumes, pvc)
+	}
```
croomes: I'm not sure about this, but would it make more sense to move this block as the "else" to the conditional at statefulset.goL36?  At the moment both sides of the conditional are in different places.

darkowlzz: Yes, that would be much better.

darkowlzz:
Forgot to add the new test file `pkg/nfs/statefulset_test.go` initially. Updated the first commit with the tests.

pkg/nfs/statefulset.go
```diff
@@ -23,16 +22,29 @@ func (d *Deployment) createStatefulSet(size *resource.Quantity, nfsPort int, htt
...
+	} else {
+		// If a PVC is provided in the NFSServer CR, add a reference to the PVC
+		// in the volumes list.
+		pvc := corev1.Volume{
+			Name: PVCNamePrefix,
+			VolumeSource: corev1.VolumeSource{
+				PersistentVolumeClaim: &d.nfsServer.Spec.PersistentVolumeClaim,
+			},
+		}
+		spec.Template.Spec.Volumes = append(spec.Template.Spec.Volumes, pvc)
```
croomes: yeah, much better

PR #173, opened 2019-09-26
NFS: cluster selection when only one cluster

darkowlzz:
When there's only one storageos cluster, select it as the current
cluster.
When there are multiple clusters, select the cluster based on
their status because the cluster controller won't allow running more
than one cluster.

### DIFF
--- a/pkg/controller/nfsserver/nfsserver_controller_test.go
+++ b/pkg/controller/nfsserver/nfsserver_controller_test.go
@@ -1,0 +1,105 @@
+package nfsserver
+
+import (
+	"reflect"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


pkg/controller/nfsserver/nfsserver_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	for _, tc := range testcases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			// Create fake storageos client.
+			stosClient := fakeStosClientset.NewSimpleClientset()
+
+			// Create the clusters.
+			for _, c := range tc.clusters {
+				stosClient.StorageosV1().StorageOSClusters("").Create(c)
```
darkowlzz: Need to add an error check here.

PR #175, opened 2019-09-30
Allow setting cluster-wide NFS container image

darkowlzz:
This adds NFSContainer in StorageOSCluster.Spec.Images to allow setting the
default NFS container image for a StorageOS cluster. This can be
overridden by specifying a container image in the NFSServer CR.

Also, adds tests for updateSpec() in the NFS controller to test the
inheritance and override of container image and storageclass from
StorageOSCluster into NFS Server.

### DIFF
--- a/pkg/apis/storageos/v1/storageoscluster_types.go
+++ b/pkg/apis/storageos/v1/storageoscluster_types.go
@@ -40,6 +40,7 @@ const (
 	CSIv0DriverRegistrarContainerImage        = "quay.io/k8scsi/driver-registrar:v0.4.2"
 	CSIv0ExternalProvisionerContainerImage    = "storageos/csi-provisioner:v0.4.2"
 	CSIv0ExternalAttacherContainerImage       = "quay.io/k8scsi/csi-attacher:v0.4.2"
+	DefaultNFSContainerImage                  = "storageos/nfs:test"
 
 	DefaultHyperkubeContainerRegistry = "gcr.io/google_containers/hyperkube"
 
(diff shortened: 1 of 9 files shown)


pkg/apis/storageos/v1/storageoscluster_types.go
```diff
@@ -40,6 +40,7 @@ const (
 	CSIv0DriverRegistrarContainerImage        = "quay.io/k8scsi/driver-registrar:v0.4.2"
 	CSIv0ExternalProvisionerContainerImage    = "storageos/csi-provisioner:v0.4.2"
 	CSIv0ExternalAttacherContainerImage       = "quay.io/k8scsi/csi-attacher:v0.4.2"
+	DefaultNFSContainerImage                  = "storageos/nfs:test"
```
croomes: Is this a good default or should we use `latest`?

pkg/controller/nfsserver/nfsserver_controller_test.go
```diff
@@ -103,3 +125,118 @@ func TestGetCurrentStorageOSCluster(t *testing.T) {
...
+					t.Fatalf("failed to get NFS Server: %v", err)
+				}
+
+				if !reflect.DeepEqual(nfsServer, tc.wantNFSServer) {
+					t.Errorf("unexpected NFS Server:\n\t(WNT) %v\n\t(GOT) %v", tc.wantNFSServer, nfsServer)
+				}
+			}
+		})
+	}
+}
```
croomes: Thanks for the tests!

pkg/apis/storageos/v1/storageoscluster_types.go
darkowlzz: It's just for now. Will set it to some proper version before the next release.
I don't see `latest` tag in docker hub. Maybe we can have some pre-release tag for the next release.

PR #178, opened 2019-10-05
e2e: Run operator locally with go run

darkowlzz:
This runs the e2e operator-sdk tests (non-OLM) with go run, instead of
running in a container.
Container image is built only for OLM bases test setup.
Also, uses the new kind load command to upload the container image from
host to a KinD environment.


darkowlzz:
There may be an issue when running the operator locally which results in the node-controller failing to update the node labels.

croomes:
LGTM but failing test?

darkowlzz:
When the operator is run this way, it's operating from outside the KinD environment. The operator controllers cannot connect to the storageos server running inside the KinD environment. This works fine for openshift test job as openshift is deployed directly on the host. A workaround for this would be to create a proxy to storageos server inside KinD from the host. We already do something similar in the e2e test for node controller tests. We can move the proxy creation before running any test.

PR #183, opened 2019-10-17
create etcd secret of the same type as source

darkowlzz:
This fixes a bug introduced in the refactoring of k8s related code due
to which the etcd secret created for storageos cluster based on a
provided reference secret is of type TLS and not the same type as the
source secret. This results in secret creation failure because TLS
secrets have a strict data field requirement and the source secret
does not have the same data fields.

Adds a test to ensure such issues are caught in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ lint:
 	go vet ./...
 
 unittest:
-	go test -v -race `go list -v ./... | grep -v test/e2e | grep -v test/olm`
+	go test -v -race `go list -v ./... | grep -v test/e2e | grep -v olm`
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -53,7 +53,7 @@ lint:
 	go vet ./...
 
 unittest:
-	go test -v -race `go list -v ./... | grep -v test/e2e | grep -v test/olm`
+	go test -v -race `go list -v ./... | grep -v test/e2e | grep -v olm`
```
darkowlzz: olm dir was changed recently.

PR #184, opened 2019-10-22
Add pod scheduler mutating admission controller

darkowlzz:
This adds a mutating admission controller for pod scheduler.
When enabled, any pod that uses a volume managed by StorageOS will be mutated to use the StorageOS scheduler.

Adds new packages `pkg/admission` and `pkg/admission/scheduler` under `internal/` for now. They provide helpers and references for adding more admission controllers in the future, if needed.

Adds e2e test to verify that the mutation works.

Refer: docs/pod-scheduler-mutating-admission-controller.md

kubebuilder webhook docs: https://book-v1.book.kubebuilder.io/beyond_basics/what_is_a_webhook.html

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -4,18 +4,48 @@ import (
 	"flag"
 	"os"
 	"runtime"
+	"strconv"
 
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
+	"github.com/storageos/cluster-operator/internal/pkg/admission"
+	"github.com/storageos/cluster-operator/internal/pkg/admission/scheduler"
 	"github.com/storageos/cluster-operator/pkg/apis"
 	"github.com/storageos/cluster-operator/pkg/controller"
+	"github.com/storageos/cluster-operator/pkg/storageos"
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	corev1 "k8s.io/api/core/v1"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+	webhookAdmission "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
 var log = logf.Log.WithName("storageos.setup")
+
+const (
+	// Env vars.
+	enableSchedulerEnvVar = "ENABLE_SCHEDULER"
+	podNamespaceEnvVar    = "POD_NAMESPACE"
+
+	// operatorNameLabel is the "name" label in the operator's deployment
+	// config. This is needed for proper operator pod selection by webhook
+	// service.
+	// NOTE: If the operator's deployment changes the name label, this must be
+	// updated.
+	operatorNameLabel = "storageos-cluster-operator"
+
+	// podSchedulerWebhookName is a fully qualified name of the pod scheduler
+	// admission webhook.
+	podSchedulerWebhookName = "podscheduler.storageos.com"
+	// podSchedulerResourceName is the name of webhook server, service,
+	// mutatingwebhookconfig and other related resources.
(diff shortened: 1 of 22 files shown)


cmd/manager/main.go
```diff
@@ -4,19 +4,53 @@ import (
...
+
+	// podSchedulerWebhookName is a fully qualified name of the pod scheduler
+	// admission webhook.
+	podSchedulerWebhookName = "podscheduler.storageos.com"
+	// podSchedulerResourceName is the name of webhook server, service,
+	// mutatingwebhookconfig and other related resources.
+	podSchedulerResourceName = "pod-scheduler-webhook"
+	// podSchedulerWebhookPort is the port at which the pod scheduler webhook
+	// server runs.
+	podSchedulerWebhookPort = 9876
```
darkowlzz: This is not the final port number. We should set it to something appropriate.

darkowlzz:
Operator deployment on openshift seems to be failing. Maybe related to permission or unsupported feature.

cmd/manager/main.go
```diff
@@ -4,19 +4,53 @@ import (
...
+	// NOTE: If the operator's deployment changes the name label, this must be
+	// updated.
+	operatorNameLabel = "storageos-cluster-operator"
+
+	// podSchedulerWebhookName is a fully qualified name of the pod scheduler
+	// admission webhook.
+	podSchedulerWebhookName = "podscheduler.storageos.com"
+	// podSchedulerResourceName is the name of webhook server, service,
+	// mutatingwebhookconfig and other related resources.
+	podSchedulerResourceName = "pod-scheduler-webhook"
```
darkowlzz: I think it'll be better to have storageos name in this. Will make it easy to identify the resources that belong to storageos.
Maybe `storageos-scheduler-webhook`.

cmd/manager/main.go
darkowlzz: Will use `5720`. Decided after internal discussion.

deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
```diff
@@ -346,9 +352,17 @@ spec:
...
                   value: cluster-operator
+                - name: POD_NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
+                - name: ENABLE_SCHEDULER
+                  value: "true"
                 ports:
-                - containerPort: 8080
+                - containerPort: 60000
```
darkowlzz: I think this is not right. Tried to make it same as in the operator deployment file. As per https://www.iana.org/form/ports-services this will be out of range `1024 - 49151`. Will set this back to 8080 for now. But it's unused at the moment. It was part of the deployment config generated by the operator-sdk initially.

internal/pkg/admission/scheduler/handler.go
```diff
@@ -0,0 +1,105 @@
...
+// Handle handles an admission request and mutates a pod object in the request.
+func (p *PodSchedulerSetter) Handle(ctx context.Context, req types.Request) types.Response {
+	// Decode the pod in request to a pod variable.
+	pod := &corev1.Pod{}
+	if err := p.decoder.Decode(req, pod); err != nil {
+		return admission.ErrorResponse(http.StatusBadRequest, err)
+	}
+	// Create a copy of the pod to mutate.
+	copy := pod.DeepCopy()
+	if err := p.mutatePodsFn(ctx, copy); err != nil {
```
domodwyer: Super clean 👍

cmd/manager/main.go
```diff
@@ -58,6 +91,55 @@ func main() {
 		fatal(err)
 	}
 
+	// Check if storageos scheduler should be enabled.
+	enableScheduler := false
```
croomes: Rather than force the var to be set, should we set the default to enabled and have an optional `DISABLE_SCHEDULER` instead?  It's good to expose options, but I think the defaults should be the recommended config? 

In this case it would also match the CR's `disableScheduler` field.

darkowlzz: Yes, that would be better, else it can become confusing.

darkowlzz: Changing the env var name to `DISABLE_SCHEDULER_WEBHOOK` to make it more clear. Not to confuse with the actual scheduler.

croomes: Thanks Sunny, LGTM!

PR #188, opened 2019-10-30
pkg/nfs: Delete NFS Server RWO volume 

darkowlzz:
With this change, when the NFS Server is deleted, the RWO PVC is also
deleted. Reclaim policy in NFSServer spec is not respected.
StorageClass reclaim policy should be used to set NFS reclaim policy.
Existing PVC used with the NFS Server are not deleted.
Also, removes the volume claim template from StatefulSet and PVC is
created separately to be used with the NFS Server. This results in
predictable PVC names which facilitates better cleanup.

Adds PersistentVolumeClaim support to k8s resource manager.


darkowlzz:
Updated the PR to not respect reclaim policy in NFS Server spec. StorageClass should be used to set reclaim policy. NFS Server spec reclaim policy is ignored if set.

PR #191, opened 2019-11-11
Add support for Prometheus ServiceMonitor

darkowlzz:
This adds support for Prometheus ServiceMonitor for NFS Server metrics.
It creates a separate service for the NFS Server for metrics endpoint, in addition to the nfs server endpoint.
The NFS Server controller, while provisioning a NFS Server, checks if `servicemonitors.monitoring.coreos.com` CRD is defined in the cluster. If the ServiceMonitor is known to the cluster, a ServiceMonitor for NFS Server metrics service is created.
A dedicated service for metrics help avoid including non-metrics endpoint in the prometheus metrics target endpoints.

The two NFS services are labelled with `app.kubernetes.io/component` to set appropriate component names. This helps in defining label selector for the ServiceMonitor that applies to the metrics service only.

Docs: docs/nfs-prometheus-metrics/README.md

Adds e2e test to check if the ServiceMonitor is created when the ServiceMonitor CRD is known to the cluster. ServiceMonitor CRD is added in `deploy/crds/` to be included in the operator-sdk e2e tests. The test also uses dynamic client to query ServiceMonitor by adding the ServiceMonitor scheme to the test framework scheme.


darkowlzz:
Code generation fails when a third-party CRD is placed in `deploy/crds/`. Need to find another location that can be used by the operator-sdk e2e test.
```
$ make generate
./build/operator-sdk generate k8s
INFO[0000] Running deepcopy code-generation for Custom Resource group versions: [storageos:[v1], ] 
INFO[0001] Code-generation complete.                    
./build/operator-sdk generate openapi
INFO[0000] Running OpenAPI code-generation for Custom Resource group versions: [storageos:[v1], ] 
Error: no API exists for Group monitoring Version v1 Kind ServiceMonitor
```

darkowlzz:
For reference, when a Service with non-metrics endpoint is selected by a ServiceMonitor, the prometheus targets page shows:
target-non-metrics-port-in-service

With a metrics only service:
single-target-metrics-endpoint

Sample graph from the collected metrics:
prometheus-servicemonitor-graph

test/e2e/util/nfs_cluster.go
```diff
@@ -103,6 +105,18 @@ func NFSServerTest(t *testing.T, ctx *framework.TestCtx) {
...
+	// ServiceMonitor is only created when the ServiceMonitor CRD is known in
+	// the cluster.
+	serviceMonitor := &monitoringv1.ServiceMonitor{}
+	smNSName := types.NamespacedName{
+		Name:      fmt.Sprintf("%s-%s", testNFSServer.Name, "metrics"),
+		Namespace: defaultNS,
+	}
+	if err := f.Client.Get(goctx.TODO(), smNSName, serviceMonitor); err != nil {
+		t.Error("failed to get NFS metrics ServiceMonitor", err)
+	}
```
darkowlzz: This is an example of using a dynamic client for third party custom resource, instead of importing a generated clientset. We can do the same for the consumers of our operator's CRDs and get rid of the clientset generation part.

pkg/util/k8s/k8s.go
```diff
@@ -102,8 +102,15 @@ func (r ResourceManager) Secret(name, namespace string, secType corev1.SecretTyp
...
-func (r ResourceManager) Service(name, namespace string, annotations map[string]string, spec *corev1.ServiceSpec) *resource.Service {
-	return resource.NewService(r.client, name, namespace, r.labels, annotations, spec)
+func (r ResourceManager) Service(name, namespace string, labels map[string]string, annotations map[string]string, spec *corev1.ServiceSpec) *resource.Service {
+	// Combine the common labels and resource specific labels.
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	for k, v := range r.labels {
+		labels[k] = v
+	}
```
darkowlzz: Introduced label passing for Service only here. Will add it to other resources as well in separate PR in order to support recommended labels for various resources https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ .

test/e2e/util/nfs_cluster.go
croomes: Interesting... that would work in the places where we've used the clientset.

PR #193, opened 2019-11-14
nfs: use k8s rest config from controller manager

darkowlzz:
Using `rest.InClusterConfig()` to get kubeconfig results in reading the
config file in the container every time the reconciliation loop runs.
File reads can be avoided by passing the k8s rest client config from
the controller manager to NFS reconciler. NFS reconciler uses the rest
client config in NFS Server deployment to create a k8s discovery client
to check if `ServiceMonitor` custom resource is defined in the cluster.


darkowlzz:
Ignore the `travis-ci/push` build. I accidentally pushed the branch to this repo, instead of my fork, and that created that build. It's failing now because I've deleted that branch from this repo and the build can't pull that branch.

PR #195, opened 2019-11-14
Scheduler must handle PVC with SC set as annotation

croomes:
In `1.5.0`, PVCs that were created using the beta annotation for specifying the StorageClass weren't being matched, causing processing of the volume to fail:

```
Internal error occurred: admission webhook "podscheduler.storageos.com" denied the request: failed to determine if the volume is managed: could not get StorageClass name associated with PVC "fast0001"
```

### DIFF
--- a/internal/pkg/admission/scheduler/handler_test.go
+++ b/internal/pkg/admission/scheduler/handler_test.go
@@ -231,7 +231,7 @@ func createPVC(name, namespace, storageClassName string, betaAnnotation bool) *c
 	}
 
 	if betaAnnotation {
-		pvc.ObjectMeta.Annotations["volume.beta.kubernetes.io/storage-class"] = storageClassName
+		pvc.ObjectMeta.Annotations[pvcStorageClassKey] = storageClassName
 	} else {
 		pvc.Spec.StorageClassName = &storageClassName
 	}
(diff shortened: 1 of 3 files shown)


internal/pkg/admission/scheduler/helper_test.go
```diff
@@ -0,0 +1,139 @@
+package scheduler
```
darkowlzz: This file already exists here https://github.com/storageos/cluster-operator/blob/master/internal/pkg/admission/scheduler/handler_test.go.
Why is it shown as a new file here?

darkowlzz: oops, never mind :sweat_smile:

internal/pkg/admission/scheduler/handler_test.go
```diff
@@ -211,22 +211,30 @@ func TestMutatePodFn(t *testing.T) {
...
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
 					corev1.ResourceStorage: resource.MustParse("1Gi"),
 				},
 			},
 		},
 	}
+
+	if betaAnnotation {
+		pvc.ObjectMeta.Annotations["volume.beta.kubernetes.io/storage-class"] = storageClassName
```
darkowlzz: Can use `pvcStorageClassKey` here.

PR #196, opened 2019-11-14
Add prometheus metrics for operator

darkowlzz:
With this change, the operator will create a metrics Service and prometheus
ServiceMonitor (when servicemonitor CRD is defined in the cluster), and
expose metrics of the CRDs and the operator itself.
Also updates the OLM operator capability to `Deep Insights`.

Refer:
- Usage docs - docs/operator-prometheus-metrics/README.md
- [Operator SDK metrics](https://github.com/operator-framework/operator-sdk/blob/master/doc/user/metrics/README.md)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -3,11 +3,15 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"os"
 	"runtime"
 	"strconv"
 
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
+	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/storageos/cluster-operator/internal/pkg/admission"
 	"github.com/storageos/cluster-operator/internal/pkg/admission/scheduler"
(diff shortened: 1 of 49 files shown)


cmd/manager/main.go
```diff
@@ -50,6 +57,12 @@ const (
 	podSchedulerAnnotationKey = "storageos.com/scheduler"
 )
 
+var (
+	metricsHost               = "0.0.0.0"
+	metricsPort         int32 = 8383
+	operatorMetricsPort int32 = 8686
```
darkowlzz: These are the default ports for operator-sdk generated projects. We can change it to something more suitable for us.

deploy/olm/storageos/storageos.clusterserviceversion.yaml
```diff
@@ -358,16 +366,18 @@ spec:
                     fieldRef:
                       fieldPath: metadata.annotations['olm.targetNamespaces']
                 - name: OPERATOR_NAME
-                  value: cluster-operator
+                  value: storageos-cluster-operator
```
croomes: Does this affect upgrades?  Do we need to test anything special?

darkowlzz: No, this shouldn't affect upgrades. It's just env var change. Renaming the deployment resource would have some affect on upgrades.

darkowlzz: This name change is required because the metrics Service uses operator's name as a label selector and it's not configurable. Setting it equal to the `name` label of the operator Deployment makes it work.

cmd/manager/main.go
```diff
@@ -140,6 +156,43 @@ func main() {
...
+
+	operatorNs, err := k8sutil.GetOperatorNamespace()
+	if err != nil {
+		fatal(err)
+	}
+
+	// CreateServiceMonitors will automatically create the prometheus-operator
+	// ServiceMonitor resources necessary to configure Prometheus to scrape
+	// metrics from this operator.
+	services := []*corev1.Service{service}
```
domodwyer: Is it OK to call this with a nil `service`?

It looks like it'll wind up creating a `v1.Service` with `Ports` set to `ServicePort{nil}` - perhaps this block should be inside an else?

darkowlzz: I took this code from operator-sdk's default generated project code.
Yes, we can move it to an else of the above service creation error check block.

PR #197, opened 2019-11-15
Add k8sDistro and prerequisites for OLM CSV

darkowlzz:
RHEL OLM CSV should have k8sDistro set to "openshift" in the example CR.

### DIFF
--- a/deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
+++ b/deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
@@ -25,6 +25,7 @@ metadata:
             "namespace": "kube-system",
             "secretRefName": "storageos-api",
             "secretRefNamespace": "default",
+            "k8sDistro": "openshift",
             "csi": {
               "enable": true,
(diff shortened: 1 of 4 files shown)


darkowlzz:
Prerequisites in OLM preview:

image

deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
```diff
@@ -25,6 +25,7 @@ metadata:
             "namespace": "kube-system",
             "secretRefName": "storageos-api",
             "secretRefNamespace": "default",
```
Arau: In the public documentation (https://docs.storageos.com/docs/platforms/openshift/install/4.1) example to install in OpenShift 4, the Secret is created in namespace `openshift-operators`, which it actually makes sense from a usability point of view. I would say that uniformity here is going to make things easier. 
The bug report also suggests to have uniformity on the CR secret reference namespace. 
https://bugzilla.redhat.com/show_bug.cgi?id=1769641.

PR #200, opened 2019-11-20
Refactor cluster controller

darkowlzz:
- Add pending cluster status when the requirements are unmet
- Add creating cluster status when the provisioning has started
- Add terminating cluster status when cluster delete is initiated
- Immediate retry for clusters that can't be provisioned because of
existing main cluster.
- Add finalizer to simplify deployment cleanup.

Previously, the phase used to be empty/unset. This resulted in the CR status to remain empty until the resource is ready. OLM scorecard test expects the CR status to not be empty. CR status will be set to pending immediately with this change.

```
$ kubectl get stos --all-namespaces 
NAMESPACE   NAME                        READY   STATUS    AGE
default     example-storageoscluster            Pending   3s

$ kubectl get stos --all-namespaces 
NAMESPACE   NAME                        READY   STATUS     AGE
default     example-storageoscluster    0/1     Creating   34s

$ kubectl get stos --all-namespaces 
NAMESPACE   NAME                        READY   STATUS    AGE
default     example-storageoscluster    1/1     Running   70s
```

Multiple instances of cluster:
```
$ kubectl get stos --all-namespaces 
NAMESPACE   NAME                        READY   STATUS    AGE
default     example-storageoscluster2           Pending   21s
olm         example                     1/1     Running   2m48s
```

### DIFF
--- a/pkg/controller/storageoscluster/storageoscluster_controller.go
+++ b/pkg/controller/storageoscluster/storageoscluster_controller.go
@@ -6,9 +6,6 @@ import (
 	"strings"
 	"time"
 
-	storageosv1 "github.com/storageos/cluster-operator/pkg/apis/storageos/v1"
-	"github.com/storageos/cluster-operator/pkg/storageos"
-	"github.com/storageos/cluster-operator/pkg/util/k8sutil"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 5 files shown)


pkg/controller/storageoscluster/storageoscluster_controller.go
```diff
@@ -110,42 +124,91 @@ func (r *ReconcileStorageOSCluster) Reconcile(request reconcile.Request) (reconc
...
+	if err := r.UpdateCurrentCluster(instance); err != nil {
+		log.Info("Failed to update current cluster", "error", err)
+		// Failed to determine or set current cluster, requeue the request.
+		return immediateRetryResult, nil
+	}
+
+	// Check if the cluster instance is marked to be deleted, which is indicated
+	// by the deletion timestamp being set.
+	isClusterToBeDeleted := instance.GetDeletionTimestamp() != nil
+	if isClusterToBeDeleted {
```
croomes: I think this is simpler?
```
// Check if the cluster instance is marked to be deleted, which is indicated
// by the deletion timestamp being set.
if instance.GetDeletionTimestamp() != nil {
...
```

PR #201, opened 2019-11-21
When getting node ip address, prefer InternalIP

croomes:
On Digital Ocean managed k8s, nodes have multiple addresses:
```
apiVersion: v1
kind: Node
...
status:
  addresses:
  - address: simon-a504
    type: Hostname
  - address: 10.138.106.43
    type: InternalIP
  - address: 159.65.74.224
    type: ExternalIP
```
The join token is being determined by joining the first address in the list, rather than the InternalIP:
```
time="2019-11-20T20:04:06.665279505Z" level=info msg=starting address=10.138.106.18 fencing=true hostname=simon-a50i id=72813024-f875-19f3-dc68-06d4d6bd853a join="simon-a50h,simon-a50i,simon-a504" module=command version="StorageOS 1.5.0 (09919f8), built: 2019-11-12T141956Z"
```
Nodes can't join the cluster as the hostname doesn't resolve:
```
time="2019-11-21T01:18:55.830711189Z" level=error msg="failed to join existing cluster" action=create category=etcd endpoint="simon-a50h,simon-a50i,simon-a504" error="Get http://simon-a504:5705/v1/members: dial tcp: lookup simon-a504 on 10.245.0.10:53: no such host" module=cp
```
This PR will use the InternalIP if defined, or use the current behaviour (first address in list) if not defined.

### DIFF
--- a/pkg/storageos/deploy.go
+++ b/pkg/storageos/deploy.go
@@ -351,11 +351,38 @@ func getPodNames(pods []corev1.Pod) []string {
 
 // GetNodeIPs returns a slice of IPs, given a slice of nodes.
 func GetNodeIPs(nodes []corev1.Node) []string {
-	var ips []string
+	ips := []string{}
 	for _, node := range nodes {
-		ips = append(ips, node.Status.Addresses[0].Address)
+		// Prefer InternalIP
+		if internalIP := GetNodeInternalIP(node.Status.Addresses); internalIP != "" {
+			ips = append(ips, internalIP)
+			continue
+		}
+		// Otherwise use first in list.
+		if address := GetFirstAddress(node.Status.Addresses); address != "" {
+			ips = append(ips, address)
+			continue
+		}
 	}
 	return ips
+}
+
+// GetNodeInternalIP the InternaIP from a slice of addresses, if it exists.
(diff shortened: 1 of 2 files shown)


pkg/storageos/deploy.go
```diff
@@ -351,13 +351,40 @@ func getPodNames(pods []corev1.Pod) []string {
 
 // GetNodeIPs returns a slice of IPs, given a slice of nodes.
 func GetNodeIPs(nodes []corev1.Node) []string {
-	var ips []string
+	ips := []string{}
 	for _, node := range nodes {
-		ips = append(ips, node.Status.Addresses[0].Address)
+		// Prefer InternalIP
+		if internalIP := GetNodeInternalIP(node.Status.Addresses); internalIP != "" {
```
domodwyer: Is there any chance of this change breaking other (non-digitalocean) setups?

croomes: I don't think so since InternalIP should always be set and used.  If for some reason InternalIP isn't set then we fall back to the previous behaviour (which probably wouldn't have worked anyways).

I've seen some k8s issues where multiple InternalIPs are set but this keeps the same behaviour (uses first).

PR #206, opened 2019-11-25
e2e: run full tests when platform supports it

darkowlzz:
This change adds platform version checks in the e2e tests. Admission
controller and NFS Server tests will be run completely on the supported
platforms and partially or skipped for unsupported platform versions.

### DIFF
--- a/pkg/storageos/scheduler_extender.go
+++ b/pkg/storageos/scheduler_extender.go
@@ -70,6 +70,9 @@ func (s Deployment) createSchedulerDeployment(replicas int32) error {
 		},
 	}
 
+	// Add cluster config tolerations.
+	s.addTolerations(&spec.Template.Spec)
+
 	// Add pod toleration for quick recovery on node failure.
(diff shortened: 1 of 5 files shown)


pkg/storageos/scheduler_extender.go
```diff
@@ -70,6 +70,9 @@ func (s Deployment) createSchedulerDeployment(replicas int32) error {
 		},
 	}
 
+	// Add cluster config tolerations.
+	s.addTolerations(&spec.Template.Spec)
```
darkowlzz: This adds the StorageOSCluster toleration to the kube-scheduler deployment. Since the k8s node in e2e test is tainted, without this toleration in kube-scheduler deployment, scheduling of kube-scheduler pod fails. Since the admission controller mutates any StorageOS volume consumer pod to use the scheduler and the scheduler isn't running, the consumer pods, including NFS Server pods fail.

PR #212, opened 2019-12-05
Openshift secret ns

croomes:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1769641

Also updates basic licence capacity in product description.

### DIFF
--- a/deploy/olm/rhel-changes.yaml
+++ b/deploy/olm/rhel-changes.yaml
@@ -15,7 +15,7 @@ metadata.annotations.alm-examples: |-
       "spec": {
         "namespace": "kube-system",
         "secretRefName": "storageos-api",
-        "secretRefNamespace": "default",
+        "secretRefNamespace": "openshift-operators",
         "k8sDistro": "openshift",
         "csi": {
           "enable": true,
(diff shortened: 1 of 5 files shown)


deploy/olm/rhel-changes.yaml
```diff
@@ -15,7 +15,7 @@ metadata.annotations.alm-examples: |-
       "spec": {
         "namespace": "kube-system",
         "secretRefName": "storageos-api",
-        "secretRefNamespace": "default",
+        "secretRefNamespace": "openshift-operators",
```
darkowlzz: Editing this file by hand is fine but the changes will be overwritten by release-gen.sh script when it's run the next, if there's anything that doesn't match the templates in the script. Same for the two storageos.clusterserviceversion.yaml files.
I believe it's not documented anywhere how to update the CSVs. The changes should be expressed in release-gen.sh file only and when the release-gen.sh script is run before the next release, it'll update this file and the CSV files.
No harm in manually editing them for now.

scripts/release-helpers/release-gen.sh
```diff
@@ -86,7 +86,7 @@ metadata.annotations.alm-examples: |-
       "spec": {
         "namespace": "kube-system",
         "secretRefName": "storageos-api",
-        "secretRefNamespace": "default",
+        "secretRefNamespace": "openshift-operators",
```
darkowlzz: If the secret is created in `openshift-operators` NS, shouldn't we move the cluster CR to the same NS?
All the resources in the same NS would be more consistent.

PR #219, opened 2019-12-16
node-ctrl: check if a node is part of the cluster

darkowlzz:
Node controller must check if a given node instance is part of the
StorageOS cluster before attempting further reconciliation.
Failure to determine if a node is a part of the StorageOS cluster should
result in an immediate retry.


pkg/controller/node/node_controller.go
```diff
@@ -131,25 +147,19 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
 		return reconcileResult, nil
 	}
 
-	return reconcileResult, nil
+	return reconcile.Result{}, nil
```
darkowlzz: A successful reconciliation need not return a periodic update in case of node controller.

PR #224, opened 2020-01-08
Add container image env vars in deployment manifest

darkowlzz:
Operator marketplaces require providing a way to set the related image used by the operator in the operator's deployment manifest as per https://github.com/zachncst/om-env-example/blob/master/README.md . In the [examples](https://github.com/zachncst/om-env-example/blob/master/ubi-noop-go/pkg/controller/ubinoop/ubinoop_controller.go#L142-L145), default fallback image is still kept and not removed from the operator.

This change creates a new package `internal/pkg/image` and moves all the default container image constants to this packages. It also defines env vars for the default container images. The default image constants will now be used as a fallback when the env var image defaults are not set.

With this, the container images used in StorageOS deployment can be set in two different places. The operator's deployment manifest can be used to set the default images. But these defaults can be overridden by StorageOSCluster CR `spec.images` property. When none of these are set, the default fallback images are used.

Since kube-scheduler image is dynamically selected, if it is set in the env var or StorageOSCluster CR, that will be used, else, a dynamic image will be used based on the k8s version.

The latest CSV files have empty values for those image env vars at the moment. When the release-gen.sh script is run for the next release generation, they will be filled with the appropriate image env var values.

### DIFF
--- a/scripts/release-helpers/release-gen.sh
+++ b/scripts/release-helpers/release-gen.sh
@@ -58,6 +58,18 @@ metadata.namespace: placeholder
 metadata.annotations.containerImage: storageos/cluster-operator:$NEW_VERSION
 spec.version: $NEW_VERSION
 spec.install.spec.deployments[0].spec.template.spec.containers[0].image: storageos/cluster-operator:$NEW_VERSION
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[0].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[1].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[2].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[3].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[4].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[5].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[6].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[7].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[8].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[9].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[10].value:
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[11].value:
 spec.replaces: storageosoperator.v$PREV_VERSION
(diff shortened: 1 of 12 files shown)


scripts/release-helpers/release-gen.sh
```diff
@@ -139,6 +151,18 @@ metadata.annotations.alm-examples: |-
...
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[4].value: registry.connect.redhat.com/storageos/csi-provisioner:v1.4.0
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[5].value: registry.connect.redhat.com/k8scsi/csi-attacher:v1.2.1
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[6].value: registry.connect.redhat.com/k8scsi/csi-attacher:v2.0.0
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[7].value: registry.connect.redhat.com/k8scsi/livenessprobe:v1.1.0
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[8].value: registry.connect.redhat.com/k8scsi/driver-registrar:v0.4.2
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[9].value: registry.connect.redhat.com/storageos/csi-provisioner:v0.4.3
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[10].value: registry.connect.redhat.com/k8scsi/csi-attacher:v0.4.2
+spec.install.spec.deployments[0].spec.template.spec.containers[0].env[11].value: registry.connect.redhat.com/storageos/nfs:1.0.0
```
darkowlzz: These values must be updated once we have the exact image URI from the registry.

croomes: Do we need to provide defaults here or are they just documentation examples?

darkowlzz: Usually the CSVs are treated like immutable releases. So, the example in the docs should be the final form of the CSV.
If we don't have all the images in the repo, we can leave them empty for now. Empty string will result in falling back to the defaults images. Using a string template would result in attempts to pull the image.
I could not find any existing operators in operatorhub.io using these env vars for reference.

croomes: In that case, I think it would be best to leave them empty for now.  I think we should prefer the current CR approach and only use the env vars where required by the marketplace.

darkowlzz: Okay, I'll leave them empty for now.

darkowlzz:
Updated `release-gen.sh` to remove the default images.

PR #229, opened 2020-02-26
Provisioner metadata

croomes:
`--extraCreateMetadata` is required to send the PVC metadata on the create request:
https://github.com/kubernetes-csi/external-provisioner#recommended-optional-arguments


darkowlzz:
As per the CI logs, the CSI helpers are failing. My guess is that this is because the new flag is unknown to the csi-provisioner and the pod is failing due to an unknown flag. We need to conditionally add this flag, maybe for csi-provisioner v1.6+ (latest stable release is v1.5) only.

pkg/apis/storageos/v1/storageoscluster_types.go
```diff
@@ -308,12 +308,17 @@ func (s StorageOSClusterSpec) GetCSIClusterDriverRegistrarImage() string {
 }
 
 // GetCSIExternalProvisionerImage returns CSI external provisioner container image.
-func (s StorageOSClusterSpec) GetCSIExternalProvisionerImage(csiv1 bool) string {
+func (s StorageOSClusterSpec) GetCSIExternalProvisionerImage(csiv1 bool, v2 bool) string {
```
darkowlzz: `nodev2` would be a better name? Since we have a lot of images with different versions here, I feel it'd be better to change the name.

PR #230, opened 2020-03-03
Add extra scheduler perms

croomes:
Fixes errors reported by the scheduler.

### DIFF
--- a/pkg/storageos/rbac.go
+++ b/pkg/storageos/rbac.go
@@ -240,6 +240,7 @@ func (s *Deployment) createClusterRoleForScheduler() error {
 				"replicationcontrollers",
 				"pods",
 				"pods/binding",
+				"pods/status",
 				"services",
 				"endpoints",
 				"events",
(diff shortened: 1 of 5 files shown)


pkg/storageos/rbac.go
```diff
@@ -240,6 +240,7 @@ func (s *Deployment) createClusterRoleForScheduler() error {
 				"replicationcontrollers",
 				"pods",
 				"pods/binding",
+				"pods/status",
```
darkowlzz: The operator also needs this permission before granting it to other resources.
Permission in manifest deployment in [role.yaml](https://github.com/storageos/cluster-operator/blob/1.5.3/deploy/role.yaml#L70) and the OLM CSV deployment [configmap](https://github.com/storageos/cluster-operator/blob/1.5.3/deploy/storageos-operators.configmap.yaml#L874) (optionally also update all the OLM CSV files with `make metadata-update`).

pkg/storageos/rbac.go
```diff
@@ -264,7 +265,7 @@ func (s *Deployment) createClusterRoleForScheduler() error {
 		{
 			APIGroups: []string{"events.k8s.io"},
 			Resources: []string{"events"},
-			Verbs:     []string{"create"},
+			Verbs:     []string{"create", "patch"},
```
darkowlzz: This permission should also be granted to the operator first like for pod status permission.

PR #232, opened 2020-03-04
Re-add provisioner flag for CSI v0 compatibility

croomes:
The flag is ignored in csi-provisioner v.1.1.0 and above, but required for v1.0.0, which we use for k8s < 1.13.


darkowlzz:
Needs rebasing.

PR #233, opened 2020-03-12
Node label sync for v2 cluster

darkowlzz:
Implements v2 API client and uses it in the node controller to sync
the node labels.

Adds package `github.com/storageos/cluster-operator/internal/pkg/client/storageos/` consisting of both v1 and v2 client and functions to simplify operations like GetNode by name and UpdateNode.

e2e test for label sync in v2 is not enabled because that requires storageos cli for inspecting the nodes from host.

Node info:
```json
[
  {
    "id": "ccf62a31-3e7c-457f-bc46-8740c7e085f2",
    "name": "kind-worker",
    "health": "online",
    "ioAddress": "10.61.0.134:5703",
    "supervisorAddress": "10.61.0.134:5704",
    "gossipAddress": "10.61.0.134:5711",
    "clusteringAddress": "10.61.0.134:5710",
    "labels": {
      "beta.kubernetes.io/arch": "amd64",
      "beta.kubernetes.io/os": "linux",
      "hello": "world",
      "kubernetes.io/arch": "amd64",
      "kubernetes.io/hostname": "kind-worker",
      "kubernetes.io/os": "linux"
    },
    "createdAt": "2020-03-12T14:09:49Z",
    "updatedAt": "2020-03-12T14:11:30Z",
    "version": "Mg"
  },
  {
    "id": "e2b39b3e-d8d5-4ca6-8253-7cbbaee93cdb",
    "name": "kind-control-plane",
    "health": "online",
    "ioAddress": "10.61.0.133:5703",
    "supervisorAddress": "10.61.0.133:5704",
    "gossipAddress": "10.61.0.133:5711",
    "clusteringAddress": "10.61.0.133:5710",
    "labels": {
      "SoMeKeY2": "SoMeVaL2",
      "apples": "5",
      "beta.kubernetes.io/arch": "amd64",
      "beta.kubernetes.io/os": "linux",
      "foo1": "bar1",
      "kubernetes.io/arch": "amd64",
      "kubernetes.io/hostname": "kind-control-plane",
      "kubernetes.io/os": "linux",
      "node-role.kubernetes.io/master": "",
      "somekey1": "someval1"
    },
    "createdAt": "2020-03-12T14:09:40Z",
    "updatedAt": "2020-03-12T14:15:26Z",
    "version": "NA"
  }
]
```


internal/pkg/client/storageos/v2/client.go
```diff
@@ -0,0 +1,114 @@
...
+	config.Host = u.Host
+	config.UserAgent = common.UserAgent
+
+	httpc := &http.Client{
+		Timeout:   common.HTTPTimeout,
+		Transport: http.DefaultTransport,
+	}
+	config.HTTPClient = httpc
+
+	if u.Scheme == common.TLSScheme {
```
domodwyer: I'm a little uneasy about this - if the user has enabled TLS on the API interface it's to prevent information like credentials being observed as they go over the network. 

A bad actor that is in a position to observe plaintext/TLS traffic as it goes over the network is almost certainly in a position to MITM the traffic flow redirecting it to their machine where they present a bogus cert, which we would accept with `InsecureSkipVerify: true`, resulting in us sending the credentials to the bad actor anyway. Normally this is impossible because unless the bad actor had obtained a valid TLS certificate, the client would not establish the connection. I understand managing the TLS certs used by the operator is probably something we can do in the future, but we probably shouldn't work with TLS endpoints unless we're honouring the commitments to maintain the security it gives us in the meantime.  

The user could put the CA cert they're using for their API into their host bundle path - this would let the operator connect to the TLS endpoint _and_ do all the verification that is needed to make TLS secure without any additional code (the HTTP client loads the host's roots by default).

internal/pkg/client/storageos/common/constant.go
```diff
@@ -0,0 +1,20 @@
...
+
+const (
+	// APIUsernameKey is the name of the k8s secret key that stores StorageOS
+	// username.
+	APIUsernameKey = "apiUsername"
+	// APIPasswordKey is the name of the k8s secret key that stores StorageOS
+	// password.
+	APIPasswordKey = "apiPassword"
+	// UserAgent is the user-agent name of the StorageOS client.
+	UserAgent = "cluster-operator"
```
domodwyer: Would it be possible to include the operator version in the user agent? It would be really helpful!

darkowlzz: Yes, that would be nice. Will have to add it to release automation.

internal/pkg/client/storageos/v2/client.go
darkowlzz: Thanks for pointing that out. We can disable this for now and enable it when we are ready for it in the future.

darkowlzz:
> Just wondering how retries work - if the node update fails, does it get retried?

Yes, returning `reconcile.Result{}` (empty), ends the reconciliation loop for a given request. Returning the `Result` with a `RequeueAfter` time, results in a retry of the same reconciliation function for the same request after the re-queue time, that's what we do by returning  variable `reconcileResult`.

darkowlzz:
Added version in the user-agent string `"cluster-operator/1.5.3"` for now. It'll be updated automatically when preparing a new release with release-gen.sh script.

domodwyer:
Looks awesome, thanks Sunny :D

PR #235, opened 2020-03-16
label-sync-test: Add deferred cleanup

darkowlzz:
Adds deferred cleanup for the test that ensures the changes are cleaned
up even when the test fails.
Adds variables for the values of labels, script files and expected
script result.


darkowlzz:
I was trying to check if the test fails if the label sync fails and it does. Still not sure how it couldn't catch the v1 label sync issue.

PR #242, opened 2020-05-05
StorageOS containers should run in kube-system by default

croomes:
Previously we advised customers to install the StorageOS containers into the `kube-system` namespace so that we can set the `system-node-critical` priority-class, which instructs Kubernetes to evict our pods last if there is resource contention.

This PR changes the default so that users don't have to manually set this.

The CRs for Openshift via the Marketplace have not changed and still use `openshift-operators`.

We are not intending to merge this into the v1 branch, so new clusters created by a v1 operator will not be affected.

### DIFF
--- a/deploy/crds/storageos_v1_storageoscluster_cr.yaml
+++ b/deploy/crds/storageos_v1_storageoscluster_cr.yaml
@@ -6,7 +6,7 @@ metadata:
 spec:
   secretRefName: "storageos-api"
   secretRefNamespace: "default"
-  namespace: "storageos"
+  namespace: "kube-system"
   # k8sDistro: openshift
   # storageClassName: fast
   # tlsEtcdSecretRefName:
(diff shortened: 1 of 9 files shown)


deploy/crds/storageos_v1_storageoscluster_cr.yaml
```diff
@@ -38,7 +38,7 @@ spec:
   #   DriverRequiresAttachment: "true"
   #   deploymentStrategy: "deployment"
   # service:
-  #   name: "storageos"
+  #   name: "kube-system"
```
darkowlzz, line 41: This shouldn't change. It's the storageos service name.

croomes: good catch!

darkowlzz:
Tests are passing well. The failure in travis-ci/push build is due to the occasional timeout when waiting for v2 cluster to be ready.

PR #243, opened 2020-05-05
v2.0.0 release prep

darkowlzz:
- Run release tools in docker.
- Update release-gen to not prefix `v` in the version.
- Update default node container to v2.0.0.


.gitignore
```diff
@@ -8,6 +8,7 @@ deploy/test
 storageos-operator.yaml
 cluster-operator.tar
 pf.log
+.cache
```
darkowlzz, line 11: This is mounted into the container when running go inside container and used to cache go build files.

Makefile
```diff
@@ -6,6 +6,15 @@ MACHINE = $(shell uname -m)
...
 BASE_IMAGE = storageos/base-image:0.2.1
 
+# Set the new version before running the release target.
+NEW_VERSION = v2.0.0
+
+# Variables to be used when running the make targets in a container.
+CACHE_DIR = $(shell pwd)/.cache
+PROJECT = github.com/storageos/cluster-operator
+GOARCH ?= amd64
+GO_VERSION = 1.14.2
```
darkowlzz, line 16: Will unify these with the build image variables a separate PR. This is for running all the tooling in container for now.

PR #244, opened 2020-05-05
Use semver version for operatorhub directory name

darkowlzz:
The CSV package directory name for releases are based on semver. Remove
the "v" prefix from the release tag when creating the CSV bundle directory.


domodwyer:
Looks good - might be better to use sed or bash pattern matching on the `v` so we don't accidentally chop off a number one day!

darkowlzz:
Thanks. Replaced `cut` with `sed` to remove any "v" at the beginning. Much better 🙂

darkowlzz:
Ignore the `travis-ci/push` build failure. I pushed my changes to this repo, instead of my fork, accidentally, and then I deleted that branch. The build is unable to pull the branch.

PR #247, opened 2020-05-18
Update OLM CSV for v2 and hide internal-objects

darkowlzz:
This updates the OLM CSV file to have v2 specific listing details and marks the all the CRDs except ClusterConfiguration as internal-objects so that they are hidden (refer https://docs.openshift.com/container-platform/4.4/operators/operator_sdk/osdk-generating-csvs.html#osdk-hiding-internal-objects_osdk-generating-csvs).

The minimum k8s version has been updated to k8s 1.13.0.

Updated the CSV files with `make metadata-update`.

### DIFF
--- a/deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
+++ b/deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
@@ -12,6 +12,12 @@ metadata:
     createdAt: "2020-06-24T11:36:42Z"
     support: StorageOS, Inc
     certified: "true"
+    operators.operatorframework.io/internal-objects: |-
+      [
+        "jobs.storageos.com",
+        "storageosupgrades.storageos.com",
+        "nfsservers.storageos.com"
+      ]
     alm-examples: |-
       [
         {
@@ -79,9 +85,10 @@ spec:
   description: |
     StorageOS is a cloud native, software-defined storage platform that
     transforms commodity server or cloud based disk capacity into
-    enterprise-class persistent storage for containers. StorageOS is ideal for
-    deploying databases, message busses, and other mission-critical stateful
-    solutions, where rapid recovery and fault tolerance are essential.
+    enterprise-class persistent storage for containers. StorageOS volumes
+    offer high throughput, low latency and consistent performance, and
+    are therefore ideal for deploying databases, message queues, and
+    other mission-critical stateful solutions.
 
     The StorageOS Operator installs and manages StorageOS within a cluster.
     Cluster nodes may contribute local or attached disk-based storage into a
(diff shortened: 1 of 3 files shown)


darkowlzz:
This is how the updates look in the operator listing page:

Screen Shot 2020-05-18 at 6 30 24 PM

The default StorageOSCluster configuration provided to the user:

Screen Shot 2020-05-18 at 6 24 46 PM

We can add more similar v2 specific changes.

Manually tested by deploying v2 with this and created volumes with workloads.

domodwyer:
That looks really cool!

Is it possible to link to the [prerequisites page on the docs](https://docs.storageos.com/docs/prerequisites/)? We need the gossip port to be open on the worker nodes too (and DFS, etc), so maybe it's best to point to the site so we can update it easier / centralised?

darkowlzz:
Yes, we can link to the prerequisites docs page. I think we'll add many more changes to the listing details v2. I'll ensure that we add a link to the prerequisites page.
Thanks.

Arau:
1. The prerequisites indicates that StorageOS needs the port 5705 open. If that is the case, the cluster won't work.

    We need all these:

    ```
    | Port Number | TCP/UDP   | Use                    |
    |:-----------:|:---------:|:---------------------- |
    | 5701        | tcp       | gRPC                   |
    | 5703        | tcp       | DirectFS               |
    | 5704        | tcp       | Dataplane Supervisor   |
    | 5705        | tcp       | REST API               |
    | 5711        | tcp & udp | Gossip service         |
    ```

    Also, I would reference the Prereq docs page as they are not optional steps https://docs.storageos.com/docs/prerequisites/

2. Etcd url has the fqdn of kubernetes. I would remove the `.cluster.local` to be agnostic of users k8s cluster name.
    (About the suggested Yaml when creating a CR instance)

    We have seen a customer using a different DNS zone and breaking some parts of StorageOS because we use the suffix in the URL.
    Also I would add the protocol (http/https). We support both, so it is easy to miss that bit as a user when using mTLS. Then StorageOS won't be able to authenticated with the etcd cluster.

    I would change:
    ```
     kvBackend:
        address: 'etcd-client.etcd.svc.cluster.local:2379'
    ```

    for

    ```
     kvBackend:
        address: 'http://etcd-client.etcd.svc:2379'
    ```

darkowlzz:
Update the listing and marked the internal CRDs as internal-objects to hide them from the UI.

Arau:
I installed one more time to verify outputs of text, links to docs and installation. The suggested CR is using the very minimal CR, which I believe is going to be part of the changes for v2.2 of the operator. 

I imagine that even this is merged, it will take enough time for us to release 2.2 operator before the changes are visible in the operatorhub.

darkowlzz:
These changes will be released as part of 2.2.
Thanks for all the reviews.

PR #250, opened 2020-06-10
Tooling and vendored deps updates

darkowlzz:
- Update vendored deps.
- Rename make targets and reorganize them.
- Add `dev-image` target to build a development image reusing the go build cache.
- Fix release-gen script to have the correct previous version value.
- Update olm diff checker script and related scripts to use local operator-sdk and yq.
- Fix operator local-run failure when run out of k8s by disabling service monitor and pod scheduler.


darkowlzz:
Planning to add golangci-lint as a separate PRs with all the lint fixes.

PR #251, opened 2020-06-12
Use golangci-lint and cleanup

darkowlzz:
- Use golangci-lint for linting and fix all the lint errors.
  - Commented out some unused constants and functions which will be used in the future.
  - Skip the directories that are generated or copied from upstream projects.
- Update Makefile to install golangci-lint and general cleanup.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,16 @@
+run:
+  skip-dirs:
+    - pkg/client
+    - internal/pkg/cert
+    - internal/pkg/crv01
+
+linters:
+  enable:
+    - gofmt
+    - goimports
+    - unconvert
+    - nakedret
+    - scopelint
+    - whitespace
+    # - gomnd
+    # - gocyclo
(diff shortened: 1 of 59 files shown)


.golangci.yml
```diff
@@ -0,0 +1,16 @@
...
+linters:
+  enable:
+    - gofmt
+    - goimports
+    - unconvert
+    - nakedret
+    - scopelint
+    - whitespace
+    # - gomnd
+    # - gocyclo
```
darkowlzz, line 16: Will enable these in the future.

.golangci.yml
```diff
@@ -0,0 +1,16 @@
+run:
+  skip-dirs:
+    - pkg/client
+    - internal/pkg/cert
+    - internal/pkg/crv01
+
+linters:
+  enable:
+    - gofmt
```
darkowlzz, line 9: Although goimports does what gofmt does, but it doesn't inform is a file is not formatted properly. Enabling gofmt separately to have overall file formatting issue reported.

PR #252, opened 2020-06-17
Update to operator-sdk v0.18.1 and go modules migration

darkowlzz:
- Migrate to go modules
- Update to operator-sdk v0.18.1
- Regenerate all the generated code (clientset, CRDs)
  - CRD file rename - new naming format introduced by the operator-sdk
- Remove `operator-sdk generate openapi` (deprecated)
- Update all the go code as per the updated dependencies.
- Skip unit tests that uses controller-runtime fake client. fake client is deprecated. Need to rewrite the tests using [envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.6.0/pkg/envtest?tab=doc)
- Disable operator-courier test due to validation failure against the CRDs generated by the latest version of OSDK. Ref: https://github.com/operator-framework/operator-courier/issues/163
- Update the automation scripts to refer the new CRD file names
- Update OLM metadata based on the newly generated CRDs


pkg/util/k8sutil/k8sutil.go
```diff
@@ -89,7 +90,7 @@ func (k K8SOps) EventRecorder() record.EventRecorder {
 // GetDeploymentsUsingStorageClassProvisioner returns a DeploymentList that use a given
 // StorageClass name in the PVC.
 func (k K8SOps) GetDeploymentsUsingStorageClassProvisioner(provisionerName string) (*appsv1.DeploymentList, error) {
-	deployments, err := k.client.AppsV1().Deployments("").List(metav1.ListOptions{})
+	deployments, err := k.client.AppsV1().Deployments("").List(context.TODO(), metav1.ListOptions{})
```
croomes, line 93: Looks like we should accept and pass a context in the k8sutil functions - let's do this separately though.

PR #253, opened 2020-06-22
Add default tolerations

darkowlzz:
Adds default tolerations to node daemonset, csi-helpers deployment, storageos-scheduler deployment and the operator deployment itself.

Default tolerations are combined with the user provided tolerations in cluster config. Since k8s internally applies some of the same tolerations in addition to the provided tolerations, there's no toleration override, new tolerations are appended to the list of tolerations.


darkowlzz:
TravisCI checks have passed but the results are not reported here https://travis-ci.org/github/storageos/cluster-operator/builds/700907954.

PR #254, opened 2020-06-23
Add default tolerations to node DaemonSet only

darkowlzz:
This adds a new function `addTolerationsWithDefaults()` to add the
cluster config tolerations with the default tolerations and changes
addTolerations() to only set the tolerations from the cluster config.

addTolerationsWithDefaults() is used by DaemonSet deployment to have
default tolerations in node DaemonSet only.

CSI helpers, scheduler and the operator itself will not have any
default tolerations.

### DIFF
--- a/pkg/storageos/deploy_test.go
+++ b/pkg/storageos/deploy_test.go
@@ -1018,7 +1018,8 @@ func TestDeployTolerations(t *testing.T) {
 				},
 				Spec: api.StorageOSClusterSpec{
 					CSI: api.StorageOSClusterCSI{
-						Enable: false,
+						Enable:             true,
+						DeploymentStrategy: "deployment",
 					},
 					Tolerations: tc.tolerations,
 				},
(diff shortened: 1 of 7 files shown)


pkg/storageos/deploy_test.go
```diff
@@ -1055,23 +1057,84 @@ func TestDeployTolerations(t *testing.T) {
...
+			nsName = types.NamespacedName{
+				Name:      createdCSIHelperDeployment.Name,
+				Namespace: createdCSIHelperDeployment.Namespace,
+			}
+			if err := c.Get(context.Background(), nsName, createdCSIHelperDeployment); err != nil {
+				t.Fatal("failed to get the created csi helpers deployment", err)
+			}
+			// Remove extra tolerations (unreachable and not-ready) that are
+			// added by k8s by default before comparison. These tolerations
+			// have non-nil toleration seconds attribuet.
```
croomes, line 1089: nit: typo

darkowlzz: thanks

PR #258, opened 2020-07-14
Fix skipped unit tests

darkowlzz:
- Rename sample CRD file to have version number (as per the Operator-SDK generated files).
- Use vendored deps when building.
- Update controller-runtime to v0.6.1 (patch release). Didn't help much with the failures due to fake client update.
- Update unit tests to fix the failures due to fake client updates
  - Update to controller-runtime's fake client broke the unit tests. The
failure errors were:
    - "resourceVersion can not be set for Create requests"
    - "failed to deploy cluster: secrets '' not found"

The resourceVersion error was occurring because tests with subtests were
creating the same cluster objects. The updated fake client retains the
previously created object. Doing a deepcopy of the object before
creating them in subtests fixes this issue.

The secret not found error was occuring when deploying the legacy driver
where the API secret is updated with a service address. Not sure what
changed in the fake client that's causing this. Since we no longer
support the legacy driver, changing the test to use CSI fixes the tests.


pkg/storageos/deploy_test.go
```diff
@@ -486,6 +486,7 @@ func TestCreateCSIHelper(t *testing.T) {
 	}
 }
 
+// NOTE: Unsupported legacy test.
```
darkowlzz, line 489: Will remove all the legacy tests when making CSI the default in cluster config separately.

PR #259, opened 2020-07-16
Don't create separate CSI registration mount if under /var/lib/kubelet

croomes:
This fixes an issue where if the node container got in a restart loop (e.g. due to typo in etcd endpoint), then mounts weren't getting cleaned up correctly leading to excessive entries in /proc/mounts.

We were creating separate mounts for the kubelet dir (/var/lib/kubelet) and the CSI registration dir (which now defaults to /var/lib/kubelet/plugins_registry).  On restart the registration dir was not cleanly unmounted due to it overlapping.  The fix is to only create the registration mount if it's not underneath the kubelet dir.

### DIFF
--- a/pkg/storageos/podspec.go
+++ b/pkg/storageos/podspec.go
@@ -2,6 +2,7 @@ package storageos
 
 import (
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


darkowlzz:
Rebased.

PR #263, opened 2020-07-23
Update pod scheduler service address

darkowlzz:
Remove cluster.local in the service endpoint to be compatible with
clusters with custom DNS configurations.

Also, the e2e tests have failed a few times recently due to failure in
ClusterConfiguration status check.
Cluster status is not updated immediately once the cluster is ready. The
reconciliation may take a few seconds. Update the status check to retry,
fetch an updated status and check again, with a timeout.


darkowlzz:
Rebased.

PR #277, opened 2020-09-24
Set app.kubernetes.io/component label

croomes:
Sets the app.kubernetes.io/component label differently for each component.  Previously they were all set to `cluster`.  New values are:
- `storageos-csi-helper` for CSI helpers.
- `storageos-daemonset` for DaemonSet.
- `storageos-scheduler` for the scehduler extender.

This will allow configuration of anti-affinty rules and Prometheus Service Monitors for individual components without adding additional labels.

### DIFF
--- a/pkg/controller/storageoscluster/cluster.go
+++ b/pkg/controller/storageoscluster/cluster.go
@@ -42,7 +42,8 @@ func (c *StorageOSCluster) SetDeployment(r *ReconcileStorageOSCluster) {
 	// by any label selectors.
 	labels["app"] = "storageos"
 
-	// Add default resource app labels.
+	// Add default resource app labels.  Component will be set to "cluster" by
+	// default.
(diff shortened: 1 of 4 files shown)


pkg/controller/storageoscluster/cluster.go
```diff
@@ -42,7 +42,8 @@ func (c *StorageOSCluster) SetDeployment(r *ReconcileStorageOSCluster) {
 	// by any label selectors.
 	labels["app"] = "storageos"
 
-	// Add default resource app labels.
+	// Add default resource app labels.  Component will be will be set to
```
darkowlzz, line 45: `Component will be set to`

darkowlzz (approved):
LGTM!

pkg/controller/storageoscluster/cluster.go
croomes: fixed - good spot

PR #280, opened 2020-10-06
Remove node v1 compatibility

croomes:
Previously the operator evaluated the node image tag to determine the StorageOS major version, and to configure accordingly.  Since Red Hat Marketplace rewrites image tags to hashes, evaluation using the tag is no longer possible.

Since there is no reason to deploy v1 with the latest operator we can remove the ability to do so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,12 +37,14 @@ install: true
 
 jobs:
   include:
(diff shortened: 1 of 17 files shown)


darkowlzz (approved):
Everything looks great!
Thanks.

PR #281, opened 2020-10-12
Set recommended max_pids used by the init container

croomes:
Set `RECOMMENDED_MAX_PIDS_LIMIT` to default 32,768, used by the init container to verify StorageOS requirements are met prior to starting the node container.  The init container will log warnings when the limit is lower than the effective limit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Parameter | Description | Default
 `secretRefNamespace` | Namespace of storageos secret |
 `namespace` | Namespace where storageos cluster resources are created | `storageos`
 `images.nodeContainer` | StorageOS node container image | `storageos/node:v2.2.0`
-`images.initContainer` | StorageOS init container image | `storageos/init:v2.0.0`
+`images.initContainer` | StorageOS init container image | `storageos/init:v2.1.0`
 `images.csiNodeDriverRegistrarContainer` | CSI Node Driver Registrar Container image | Varies depending on Kubernetes version
 `images.csiClusterDriverRegistrarContainer` | CSI Cluster Driver Registrar Container image |  Varies depending on Kubernetes version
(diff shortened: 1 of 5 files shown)


pkg/apis/storageos/v1/storageoscluster_types.go
```diff
@@ -202,6 +205,20 @@ type StorageOSClusterSpec struct {
...
+	// can be created by a single container.
+	//
+	// In most environments this will be at least the StorageOS recommended
+	// minimum of 32768, but the limit can be as low as 1024 in CRI-O based
+	// distributions such as Openshift.
+	//
+	// Before changing this value, consult the StorageOS documentation.  A
+	// negative value disables the check entirely.  Unset or '0' uses the
+	// StorageOS default.
+	MinimumPidLimit int `json:"minimumPidLimit,omitempty"`
```
croomes, line 221: I'm not sure we want to add to the CRD until we set to enforcing as plans may change how to do it - thoughts?

Arau: Even if the position is to have it disabled by default with -1, or using the defaults with 0, I have the impression that the CRD will need the flag anyway. And then it is the defaults in the operator that decide what to do. 
Since it has been agreed that PRE has to be able to disable it, having the CRD seems like the only option at the moment to expose that possibility, regardless of the implementation, and who executes the check/barrier.

Arau (approved):

pkg/apis/storageos/v1/storageoscluster_types.go
croomes: yes, we will almost certainly re-intoduce as-is, but for now there's no benefit to including it, so we agreed to remove.

darkowlzz (approved):
LGTM!

PR #284, opened 2020-10-23
Set scheduler when pvcs use default storageclass

croomes:
A PVC provisioned using the default StorageClass could cause the admission controller scheduler selection to fail with:
```
admission webhook "podscheduler.storageos.com" denied the request: failed to determine if the volume is managed: could not get StorageClass name associated with PVC "pvc-test"
```
This adds support for evaluating default StorageClasses.

### DIFF
--- a/test/e2e/util/admission_controller.go
+++ b/test/e2e/util/admission_controller.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"context"
 	goctx "context"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


test/e2e/util/admission_controller.go
```diff
@@ -44,79 +58,547 @@ func PodSchedulerAdmissionControllerTest(t *testing.T, ctx *framework.Context) {
...
+	if err != nil {
+		return err
+	}
+	t.Logf("deleted %s %s/%s", obj.GetObjectKind().GroupVersionKind().Group, key.Namespace, key.Name)
+	return nil
+}
+
+// DisableDefaultStorageClass disables the default StorageClass, if set.  It
+// returns a function that re-enables it.
+func DisableDefaultStorageClass(t *testing.T, fc framework.FrameworkClient) (func() error, error) {
```
darkowlzz, line 577: Great! 😄  👍

darkowlzz (approved):
LGTM!
Thanks for all the detailed tests.

PR #287, opened 2020-10-28
Advertise RWX volume feature

avestuk:
This PR should be repointed at another branch but I wasn't sure exactly what was required so we can just do the review and then create the necessary branch to PR into.


timothy-storageos (approved):

croomes (approved):

actionbuddha (commented):
Are our access modes really 'mutable' ? Multiple, perhaps, not sure we can change between them though,

actionbuddha (approved):
LGTM - thanks @avestuk

PR #294, opened 2020-11-26
e2e test reporting

croomes:
By running the subtests using t.Run() it's more clear which test may be failing as logs will be prefixed with the name / subtest name.


darkowlzz (approved):
The report looks very good now. Thanks.

PR #295, opened 2020-11-27
Set the default container for kubectl logs

croomes:
Adds `kubernetes.io/default-logs-container: storageos` annotation to the DaemonSet Pod to specify a default container to return logs from.

### DIFF
--- a/pkg/storageos/deploy.go
+++ b/pkg/storageos/deploy.go
@@ -35,7 +35,7 @@ const (
 	statefulsetKind = "statefulset"
 	deploymentKind  = "deployment"
 
-	daemonsetName   = "storageos-daemonset"
+	DaemonSetName   = "storageos-daemonset"
 	statefulsetName = "storageos-statefulset"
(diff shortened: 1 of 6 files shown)


pkg/storageos/deploy.go
```diff
@@ -35,7 +35,7 @@ const (
 	statefulsetKind = "statefulset"
 	deploymentKind  = "deployment"
 
-	daemonsetName   = "storageos-daemonset"
+	DaemonSetName   = "storageos-daemonset"
```
avestuk, line 38: Out of curiosity why does this var need to be public now?

avestuk (approved):

pkg/storageos/deploy.go
darkowlzz: It's being used in the e2e test for this change.

darkowlzz (approved):
LGTM!
Thanks for the nice tests.

PR #311, opened 2021-03-16
Toleration defaults for helper pods

croomes:
Sets default tolerations on the non-critical helper pods.  

By default, the helpers will tolerate `node.kubernetes.io/disk-pressure` and the default `TolerationSeconds` for `node.kubernetes.io/not-ready` and `node.kubernetes.io/unreachable` is reduced from 300 to 30s.

- Allows defaults to be overridden by tolerations set in the StorageOSCluster spec.  
- Removes the deprecated `node.kubernetes.io/out-of-disk` toleration.
- Sorts tolerations on key name to ensure idempotency.

### DIFF
--- a/pkg/apis/storageos/v1/storageoscluster_types.go
+++ b/pkg/apis/storageos/v1/storageoscluster_types.go
@@ -571,7 +571,7 @@ func (s StorageOSClusterSpec) GetHelperTolerations(tolerationSeconds int64) []co
 }
 
 // mergeTolerations merges a slice of tolerations on top of a base slice,
-// overwiting duplicate keys.
+// overwriting duplicate keys.
 func mergeTolerations(base []corev1.Toleration, overlay []corev1.Toleration) []corev1.Toleration {
 	tolerations := make(map[string]corev1.Toleration)
 	for _, t := range append(base, overlay...) {
--- a/pkg/apis/storageos/v1/storageoscluster_types.go
+++ b/pkg/apis/storageos/v1/storageoscluster_types.go
@@ -554,17 +554,35 @@ func (s StorageOSClusterSpec) GetCSIDeploymentStrategy() string {
 	return DefaultCSIDeploymentStrategy
 }
 
-// GetTolerations returns the tolerations applied to all the StorageOS related
-// pods.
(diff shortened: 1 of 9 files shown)


avestuk (approved):

pkg/storageos/scheduler_extender.go
```diff
@@ -66,14 +66,11 @@ func (s Deployment) createSchedulerDeployment(replicas int32) error {
 		},
 	}
 
-	// Add cluster config tolerations.
-	if err := s.addTolerations(&spec.Template.Spec); err != nil {
+	// Add helper tolerations.
+	if err := s.addHelperTolerations(&spec.Template.Spec, podTolerationSeconds); err != nil {
```
darkowlzz, line 70: Looks like `addCommonPodProperties()` is not added to the scheduler pod. It exists in api-manager and csi-helpers. We can replace `addHelperTolerations()` with `addCommonPodProperties()` to add toleration, affinity and any other common attributes.

pkg/apis/storageos/v1/storageoscluster_types.go
```diff
@@ -554,17 +554,35 @@ func (s StorageOSClusterSpec) GetCSIDeploymentStrategy() string {
...
+	t := mergeTolerations(toleration.GetDefaultHelperTolerations(tolerationSeconds), s.Tolerations)
+	toleration.Sort(t)
+	return t
+}
 
-	// Append the tolerations specified in the spec.
-	if len(s.Tolerations) > 0 {
-		tolerations = append(tolerations, s.Tolerations...)
+// mergeTolerations merges a slice of tolerations on top of a base slice,
+// overwiting duplicate keys.
```
darkowlzz, line 574:
```suggestion
// overwriting duplicate keys.
```

darkowlzz (commented):
Overall looks good to me.
Added a few inline comments.
Thanks.

pkg/storageos/scheduler_extender.go
croomes: yes, it seemed odd that the scheduler wasn't applying `addCommonPodProperties()`.  Added it.

PR #312, opened 2021-03-16
Remove priorityclass from helper pods

croomes:
Removes the `system-node-critical` priority class from the helper Pods.  This will allow the pods to failover more quickly when the node is under pressure.

### DIFF
--- a/pkg/storageos/deploy_test.go
+++ b/pkg/storageos/deploy_test.go
@@ -1527,8 +1527,7 @@ func TestDeployTLSEtcdCerts(t *testing.T) {
 }
 
 // TestDeployPodPriorityClass tests that the pod priority class is set properly
-// for the daemonset and statefulset pods when deployed in kube-system
-// namespace.
+// for the daemonset when deployed in kube-system namespace.
 func TestDeployPodPriorityClass(t *testing.T) {
 	testCases := []struct {
 		name                  string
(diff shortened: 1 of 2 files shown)


pkg/storageos/deploy_test.go
```diff
@@ -1527,31 +1527,37 @@ func TestDeployTLSEtcdCerts(t *testing.T) {
 }
 
 // TestDeployPodPriorityClass tests that the pod priority class is set properly
-// for the daemonset and statefulset pods when deployed in kube-system
-// namespace.
+// for the daemonset when deployed in kube-system namespace.
```
darkowlzz, line 1530: Because the priority class is cluster scoped and the old behavior of kube-system exclusivity is removed, maybe this whole test can be removed. We don't set priority class on anything except the daemonset and the daemonset will always have the priority class set, regardless of the namespace.

darkowlzz (commented):
Removal of priority class from common pod properties looks good to me.
I've added a comment in #311 to add the common pod properties in scheduler as well.

pkg/storageos/deploy_test.go
croomes: Since we still only set the DaemonSet priority class when it runs in `kube-system`, we should probably keep the test until we remove that restriction.  I can do that in a separate PR.  I think we still allow deploying CSI as a StatefulSet, so I kept those tests too, though I agree they're not very useful.

PR #314, opened 2021-03-23
Prep for v2.3.4

croomes:
Generate manifests and update metadata files with NEW_VERSION=v2.3.4 make release.


bundle/manifests/storageosoperator.clusterserviceversion.yaml
```diff
@@ -564,7 +567,8 @@ spec:
   minKubeVersion: 1.13.0
   provider:
     name: StorageOS, Inc
+  replaces: storageosoperator.v2.3.3
```
darkowlzz, line 570: Replaces is not needed in the new bundle format.

darkowlzz (commented):
Overall looks good to me.
There's an extra field in the bundle CSV, left an inline comment about it.

bundle/manifests/storageosoperator.clusterserviceversion.yaml
croomes: Removed - thanks

PR #315, opened 2021-04-06
Move pod scheduler webhook to api-manager

croomes:
Moves the pod scheduler mutating webhook server to the api-manager: https://github.com/storageos/api-manager/pull/40

The operator is now only responsible for creating the webhook configuration, the service and the RBAC rules required by api-manager.

Previously, `DISABLE_SCHEDULER_WEBHOOK` could be set as an environment variable on the operator to disable the scheduler and to not set up the webhook server or configuration rules.  

Since we will now have multiple handlers all requiring the webhook (there will be two: one mutating, another validating), there is no longer an option not to create the webhook configuration.  Instead, to disable the Pod scheduler, set `disableScheduler=true` in the StorageOSCluster CR.  This is the same as before and makes the pod scheduler webhook a no-op.


pkg/storageos/webhook.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+const (
+	// MutatingWebhookConfigName is the name used for the mutating webhook
+	// configuration.
+	MutatingWebhookConfigName = "storageos-mutating-webhook"
+
+	// MutatingWebhookServiceName name is the name of the service that will be
+	// created for the mutating webhook.
+	MutatingWebhookServiceName = "storageos-mutating-webhook"
```
darkowlzz, line 18: This service name will be common to the mutating webhook and validating webhooks in the future. Let's move this into a common variable with a common name, not specific to any type of webhook.

croomes: Changed to `storageos-webhook`

pkg/storageos/api_manager.go
```diff
@@ -121,6 +123,12 @@ func (s Deployment) createAPIManagerDeployment(replicas int32) error {
 		},
 	}
 
+	// If the Scheduler is disabled, set the scheduler name to an empty string.
+	// This will cause the pod scheduler webhook handler to be a no-op.
+	if strconv.FormatBool(s.stos.Spec.DisableScheduler) == "true" {
```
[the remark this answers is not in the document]
croomes: doh! yes.

pkg/storageos/deploy.go
```diff
@@ -17,9 +17,9 @@ const (
 	// IntreeProvisionerName is the name of the k8s native provisioner.
 	IntreeProvisionerName = "kubernetes.io/storageos"
 	// CSIProvisionerName is the name of the CSI provisioner.
-	CSIProvisionerName = "storageos"
+	StorageOSProvisionerName = "storageos"
 	// StorageOSProvisionerName is the new CSI provisioner name.
-	StorageOSProvisionerName = "csi.storageos.com"
+	CSIProvisionerName = "csi.storageos.com"
```
darkowlzz, line 22: The comments above the variables have been interchanged :slightly_smiling_face:

croomes: Actually, the comments were still good, the values were swapped.  Fixed now.

darkowlzz (approved):
LGTM!

PR #316, opened 2021-04-13
Fix apiManagerContainer tag in v1 deploy CRD

mhmxs:
`apiControllerContainer` was an invalid image name.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,14 +52,14 @@ jobs:
         - "INSTALL_METHOD=none"
       name: KinD (k8s-1.19)
       before_script:
-        - docker login -u "$REGISTRY_USER" -p "$REGISTRY_PASS"
+        - docker login -u "$REGISTRY_USER" -p "$REGISTRY_PASS"  || true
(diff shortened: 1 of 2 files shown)


croomes (approved):
Nicely spotted!

darkowlzz:
We added docker login in CI to avoid failures due to image pull limitation but looks like login fails for PRs created by users who aren't part of travisCI org or registered. Maybe the credentials aren't being shared.
I'm not sure of a better way to do it. Maybe attempt to login and continue even if it fails? That way, if the login fails and docker hub limits are not reached, the CI will still function.

darkowlzz:
@mhmxs can you try adding `|| true` in https://github.com/storageos/cluster-operator/blob/690bced03f2cebc89a6dd5d877b1f81964db53e4/.travis.yml#L55 and the second login in the deploy stage?

mhmxs:
> The CI keeps on waiting for some input:

Experienced the same :D

mhmxs:
Tracked here: https://github.com/storageos/cluster-operator/pull/320

PR #318, opened 2021-04-14
Add pvc mutator config

croomes:
Adds the PVC mutator configuration.


pkg/storageos/webhook.go
```diff
@@ -60,47 +65,60 @@ func (s Deployment) createWebhookService() error {
...
-		},
+		}),
 	}
 	labels := podLabelsForAPIManager(s.stos.Name)
 
 	return s.k8sResourceManager.MutatingWebhookConfiguration(MutatingWebhookConfigName, labels, webhooks).Create()
 }
 
+func (s Deployment) mutatingWebhookConfiguration(name string, port int32, path string, rules []admissionv1.RuleWithOperations) admissionv1.MutatingWebhook {
+	failurePolicy := admissionv1.Ignore
```
darkowlzz, line 99: Shouldn't this also be a function parameter to be able to have different failure policies for Pod and PVC webhook configurations?

pkg/storageos/webhook.go
```diff
@@ -60,47 +65,60 @@ func (s Deployment) createWebhookService() error {
...
-			AdmissionReviewVersions: []string{"v1"},
-		},
+		}),
 	}
 	labels := podLabelsForAPIManager(s.stos.Name)
 
 	return s.k8sResourceManager.MutatingWebhookConfiguration(MutatingWebhookConfigName, labels, webhooks).Create()
 }
 
+func (s Deployment) mutatingWebhookConfiguration(name string, port int32, path string, rules []admissionv1.RuleWithOperations) admissionv1.MutatingWebhook {
```
darkowlzz, line 98: I wonder if we need to set ports. It works fine with the default 443 and 9443. I don't see only the port changing in the future. I think there's a high possibility of the webhook service to change as well.
So, maybe add both or none for now and add them in the future when we need to change them.

mhmxs (approved):

pkg/storageos/webhook.go
croomes: Yes, I've changed it to be a function param but left the PVC webhook to ignore failures for now.  I want to do more testing before setting to fail.

pkg/storageos/webhook.go
croomes: I've removed the port for now - it's not something we support changing, so we can revisit later if needed.

pkg/storageos/webhook.go
```diff
@@ -60,47 +65,60 @@ func (s Deployment) createWebhookService() error {
...
-				{
-					Operations: []admissionv1.OperationType{admissionv1.Create},
-					Rule: admissionv1.Rule{
-						APIGroups:   []string{""},
-						APIVersions: []string{"v1"},
-						Resources:   []string{"pods"},
-						Scope:       &scopeAll,
-					},
+		}),
+		s.mutatingWebhookConfiguration(webhookPVCMutatorName, mutatePVCPath, admissionv1.Ignore, []admissionv1.RuleWithOperations{
```
darkowlzz, line 81: Didn't we decide to make the failure policy to fail for PVC mutating webhooks? https://github.com/storageos/api-manager/pull/42#discussion_r612606661

croomes: Yes, I just wanted to enable it separately after doing more testing: https://github.com/storageos/cluster-operator/pull/318#discussion_r613880899

PR #320, opened 2021-04-15
Fix apiManagerContainer tag in v1 deploy CRD

mhmxs:
`apiControllerContainer` was an invalid image name.


darkowlzz (approved):
Thanks!

croomes (approved):
Well spotted!

PR #323, opened 2021-04-20
Set webhook service-for label to be unique

croomes:
The webhook service has the `app.kubernetes.io/service-for=storageos-api-server` label set.  It should be `app.kubernetes.io/service-for=storageos-webhook-server`.

Having multiple services with the same label means you can't search for a service using label selectors, which the kubectl plugin does.

`storageos-api-server` was getting set as the `combineLabels()` function was updating the base labels unintentionally.

### DIFF
--- a/test/e2e/util/cluster.go
+++ b/test/e2e/util/cluster.go
@@ -430,6 +430,41 @@ func APIManagerDeploymentTest(t *testing.T, ns string, retryInterval, timeout ti
 	}
 }
 
+// APIManagerWebhookServiceTest checks the api-manager webhook service deployment.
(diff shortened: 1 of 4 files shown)


mhmxs (approved):

test/e2e/util/cluster.go
```diff
@@ -430,6 +430,41 @@ func APIManagerDeploymentTest(t *testing.T, ns string, retryInterval, timeout ti
 	}
 }
 
+// APIManagerWebhookServiceTest checks the api-manager webhook service deployment.
+func APIManagerWebhookServiceTest(t *testing.T, ns string, retryInterval, timeout time.Duration) {
```
darkowlzz, line 434: @croomes looks like this wasn't added in clusterCSINodeV2_test.go file to be executed.

PR #324, opened 2021-04-21
Add webhook resource migration

darkowlzz:
The webhook resource migration runs at operator startup, before the
controllers are started and ensures that the old webhook related
resources are deleted if they are found. The migration has a timeout of
20 seconds. If the migration gets stuck for some reason, the context
timeout will help cancel the migration operations and return failure.

Also adds fake k8s client based test and kind based e2e tests.

The operator-sdk local run command is updated to have valid options and
the go.sum file is updated with `make tidy`.


darkowlzz:
Docker pull rate limit hit in the CI. Closing this for a PR from the repo branch instead of fork.

PR #325, opened 2021-04-21
Add webhook resource migration

darkowlzz:
The webhook resource migration runs at operator startup, before the
controllers are started and ensures that the old webhook related
resources are deleted if they are found. The migration has a timeout of
20 seconds. If the migration gets stuck for some reason, the context
timeout will help cancel the migration operations and return failure.

Log message when some legacy resource is found and deleted:

```console
2021-04-22T18:27:35.240Z        INFO    storageos.setup.migration       deleted legacy resource {"kind": "MutatingWebhookConfiguration", "key": "/storageos-scheduler-webhook"}
2021-04-22T18:27:35.268Z        INFO    storageos.setup.migration       deleted legacy resource {"kind": "Service", "key": "storageos-operator/storageos-scheduler-webhook"}
```

Nothing is logged when no legacy resource is found.

Also adds fake k8s client based test and kind based e2e tests.

The operator-sdk local run command is updated to have valid options and
the go.sum file is updated with `make tidy`.

### DIFF
--- a/cmd/manager/migration.go
+++ b/cmd/manager/migration.go
@@ -1,0 +1,145 @@
+package main
+
+import (
+	"context"
+	"time"
+
+	"github.com/go-logr/logr"
+	oputils "github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	admissionv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/storageos/cluster-operator/pkg/util/k8s/resource"
+)
+
+// **Upgrade Migrations**
+//
+// Webhook configuration and service migration:
+//
+// Before cluster-operator v2.4.0, the admission webhook server ran in
+// the cluster-operator itself. Some old controller-runtime admission
+// controller tooling created the webhook configurations and services at
+// operator startup. These resources are unmanaged and are left behind even
+// after uninstalling the cluster-operator. In cluster-operator v2.4.0, the
+// admission webhook server is moved to the api-manager component and the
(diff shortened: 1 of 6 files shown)


cmd/manager/migration.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+
+	// Use a context with timeout to be able to exit the migration within a
+	// reasonable time if the operations get stuck.
+	ctx, cancel := context.WithTimeout(context.Background(), migrationTimeoutDuration)
+	defer cancel()
+
+	// Run migration actions.
+
```
mhmxs, line 74: Extra empty line

cmd/manager/migration.go
```diff
@@ -0,0 +1,130 @@
...
+// webhookMigration runs the webhook migration, deleting the old webhook
+// related resources.
+func webhookMigration(ctx context.Context, c client.Client, log logr.Logger, namespace string) error {
+	// Check if the known resources exist and delete them if found, no-op if
+	// not found.
+	mutatingWHC := &admissionv1.MutatingWebhookConfiguration{}
+	mhcKey := types.NamespacedName{Name: oldWebhookResourceName}
+	if err := deleteIfFound(ctx, c, mhcKey, mutatingWHC); err != nil {
+		migrationFailLog(log, resource.MutatingWebhookConfigurationKind, mhcKey.String(), err.Error())
+		return err
```
mhmxs, line 93: Doe is make sense to delete service on case of errors during delete operation of config?

mhmxs (approved):

cmd/manager/migration.go
darkowlzz: Added intentionally to create a separate section. Not strictly associated with the following webhook migration action.

cmd/manager/migration.go
darkowlzz: Yes, need to think about it. Should the deletion continue deleting others, not caring for failures. Not it's less likely to fail, if webhook config deletion fail, there must be something wrong in the cluster or the client. The following actions will also fail.

mhmxs: :%s/will/might :) I'm fine with it, just was an idea.

croomes (approved):
Thanks, just a suggestion about the logging, but fine to merge as-is if preferred.

PR #328, opened 2021-04-26
Add rbac needed for csi-resizer v1.0.0

croomes:
The CSI external-resizer v1.0.0 introduced a new RBAC requirement.  This grants read-only access to Pods.

Tested manually and with kubecover.


darkowlzz (approved):
Thanks!

PR #333, opened 2021-05-12
Set priority class for helper pods

croomes:
Sets the `system-cluster-critical` priority class on StorageOS pods that must be running somewhere in the cluster.  This allows the Pod to move to another node if it doesn't tolerate a taint (such as memory pressure) but ensures it runs somewhere.

### DIFF
--- a/pkg/storageos/podspec.go
+++ b/pkg/storageos/podspec.go
@@ -9,11 +9,11 @@ import (
 )
 
 const (
-	// Name of kube-system namespace.
-	kubeSystemNamespace = "kube-system"
+	// Name of the node critical priority class.
+	nodeCriticalPriorityClass = "system-node-critical"
 
-	// Name of the critical priority class.
-	criticalPriorityClass = "system-node-critical"
+	// Name of the cluster critical priority class.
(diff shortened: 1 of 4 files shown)


mhmxs (approved):

pkg/storageos/podspec.go
```diff
@@ -273,9 +273,24 @@ func (s *Deployment) addTLSEtcdCerts(podSpec *corev1.PodSpec) {
...
 	}
+	return nil
+}
+
+// addPodPriorityClass sets the priority class for a Pod.
+//
+// Prior to StorageOS v2.4.0, the priority class was only set for Pods running
+// in the kube-system namespace, as k8s 1.16 and earlier would only allow this.
+func (s *Deployment) addPodPriorityClass(podSpec *corev1.PodSpec, priorityClass string) {
+	podSpec.PriorityClassName = priorityClass
```
darkowlzz, line 295: Getting rid of the kube-system namespace conditional, finally :smiley: 
Maybe the comment about kube-system is not necessary.

darkowlzz (approved):
LGTM!

PR #335, opened 2021-05-13
Increase provisioner timeout from default [15s] to 30s

croomes:
In normal operation with many thousands of volumes in the cluster, provision times are well under 10s.  If we do encounter an issue that affects provisioning times (e.g. load/disk issues on etcd) it's more efficient to wait rather than retry.


darkowlzz (approved):

mhmxs (approved):

mhmxs:
Increase provisioner timeout from `default [15s]` to 30s ;)

PR #336, opened 2021-05-20
Changes to the StorageOSCluster CR get applied to StorageOS

croomes:
Previously the StorageOSCluster CR was treated as immutable.  With this change, a user can modify settings in the CR spec and those changes will get passed to StorageOS.

Some configuration items at set via an API call to the StorageOS control plane API.  Others are set in a ConfigMap that is passed to the node container as env vars.  Items that are configured via the API are also set in the ConfigMap once they have been successfully applied.

The control plane determines which configuration items can be updated at run-time, with a restart, or not at all.


pkg/util/k8s/resource/configmap.go
```diff
@@ -36,8 +36,10 @@ func NewConfigMap(c client.Client, name, namespace string, labels map[string]str
 // Get returns an existing ConfigMap and an error if any.
 func (cm ConfigMap) Get() (*corev1.ConfigMap, error) {
 	configmap := &corev1.ConfigMap{}
-	err := cm.client.Get(context.TODO(), cm.NamespacedName, configmap)
-	return configmap, err
+	if err := cm.client.Get(context.TODO(), cm.NamespacedName, configmap); err != nil {
```
mhmxs, line 39: I know it is out of the scope of the change, but are we able to replace TODO or we really don't know what to do with context here?

test/e2e/clusterCSINodeV2_test.go
```diff
@@ -62,12 +68,13 @@ func TestClusterCSINodeV2(t *testing.T) {
...
-		Namespace: namespace,
-	}
 	t.Run("ClusterStatusCheck", func(t *testing.T) {
-		if err = testutil.ClusterStatusCheck(t, namespacedName, 1, testutil.RetryInterval, testutil.Timeout); err != nil {
+		if err = testutil.ClusterStatusCheck(t, clusterNsName, 1, testutil.RetryInterval, testutil.Timeout); err != nil {
+			logs, logErr := util.GetOperatorLogs(namespace)
+			if logErr != nil {
+				t.Error(errors.Wrap(logErr, "failed to fetch operator logs"))
+			}
+			t.Log(logs)
```
mhmxs, line 77: I suggest to Log first and than check logErr, because for me looks more readable if Error and Fatal appear at the end

test/e2e/util/cluster_update.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		t.Fatalf("failed to get operator namespace: %v", err)
+	}
+	nsName := types.NamespacedName{
+		Name:      TestClusterCRName,
+		Namespace: namespace,
+	}
+
+	stos := &storageos.StorageOSCluster{}
+	if err := f.Client.Get(goctx.TODO(), nsName, stos); err != nil {
```
mhmxs, line 32: Can i ask the reason of TODOs in this file?

mhmxs (commented):
Small

pkg/storageos/configmap.go
```diff
@@ -121,19 +126,89 @@ const (
...
 )
 
-// createService creates a ConfigMap to store the node container configuration.
-func (s *Deployment) createConfigMap() error {
+// ensureConfigMap creates or updates a ConfigMap to store the node container
+// configuration.
+func (s *Deployment) ensureConfigMap() error {
 	config := configFromSpec(s.stos.Spec, CSIV1Supported(s.k8sVersion))
 
 	labels := make(map[string]string)
```
avestuk, line 134: What's going on here? 

It looks like the call to `s.k8sResourceManager.ConfigMap(configmapName, s.stos.Spec.GetResourceNS(), labels, config).Get()` calls `ResourceManager.combineLabels()` which creates a new map and combines the `ResourceManager.Labels` with the labels that are passed in. 

Would passing in nil risk causing an explosion later because we'd instantiate r with a nil map for it's labels?

pkg/storageos/configmap.go
```diff
@@ -121,19 +126,89 @@ const (
...
+	existing, err := s.k8sResourceManager.ConfigMap(configmapName, s.stos.Spec.GetResourceNS(), labels, config).Get()
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	// If there is an existing CM, check for changes and apply any that are set
+	// dynamically via the API.
+	if existing != nil {
+		// Don't update if unchanged.
+		if reflect.DeepEqual(config, existing) {
```
avestuk, line 145: How can a `map[string]string` and `*v1.ConfigMap` ever be deeply equal? Would you not need to turn the configMap into a map in order to compare them?

avestuk (approved):

pkg/util/k8s/resource/configmap.go
croomes: Yes, it would be better to pass in a proper context, but this is a wider change to do consistently.  The k8s client function signature changed and this is in keeping with the existing code.  I'd prefer to leave as-is for now and we can do a follow-on PR if needed.

test/e2e/clusterCSINodeV2_test.go
croomes: I need to leave the Fatal at the end as it exits, but I'm re-arranged to try and make it more readable.

internal/pkg/storageos/client.go
```diff
@@ -0,0 +1,167 @@
...
+	config.Host = u.Host
+	if !strings.Contains(u.Host, ":") {
+		config.Host = fmt.Sprintf("%s:%d", u.Host, DefaultPort)
+	}
+
+	httpc := &http.Client{
+		Timeout:   HTTPTimeout,
+		Transport: transport,
+	}
+	config.HTTPClient = httpc
```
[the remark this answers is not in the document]
croomes: no, well spotted!

pkg/storageos/configmap.go
darkowlzz: Should be just fine to pass nil. The checks are handled internally.

test/e2e/clusterCSINodeV2_test.go
mhmxs: perfect i just say:
```
t.Log(logs)
if logErr != nil t.Error(errors.Wrap(logErr, "failed to fetch operator logs"))
t.Fatal(err)
```
Because Error doesn't stops execution.

pkg/storageos/configmap.go
croomes: No, combineLabels checks for nil and returns r.labels, which is never nil since it's set in the constructor.  So passing in nil is ok, and it would be safe to remove `labels := make(map[string]string)`. I'd prefer to keep it as I think it makes it more clear that we don't want to set specific labels.

pkg/storageos/configmap.go
croomes: doh!  Very well spotted.  It needs to compare with `existing.Data`.  I'm adding a unit test for the function...

croomes: So that test became a bit more involved than I expected...

darkowlzz (approved):
LGTM!
A lot of detailed tests. Thanks.

PR #338, opened 2021-06-02
Increase scheduler weight

mhmxs:
In `kube-scheduler` scheduler plugins and extender schedulers are in different scale. In some cases Kubernetes schedules a pod far of the master volume, because extender's low numbers can't make any sense. This change introduces the weight of our scheduler to fix pod locality issues.


darkowlzz (approved):
LGTM! Thanks.

croomes (approved):

PR #339, opened 2021-06-03
Create roles for scheduling

mhmxs:
New implementation of scheduler uses api server to detect pod volume types, and it needs more rights to do.

### DIFF
--- a/pkg/storageos/rbac.go
+++ b/pkg/storageos/rbac.go
@@ -38,11 +38,11 @@ const (
 	NFSClusterRoleName    = "storageos:nfs-provisioner"
 	NFSClusterBindingName = "storageos:nfs-provisioner"
 
-	KubeSchedulerClusterRoleName    = "storageos:kube-scheduler"
-	KubeSchedulerClusterBindingName = "storageos:kube-scheduler"
-
 	SchedulerExtenderClusterRoleName    = "storageos:scheduler-extender"
 	SchedulerExtenderClusterBindingName = "storageos:scheduler-extender"
+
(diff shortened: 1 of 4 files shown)


pkg/storageos/rbac.go
```diff
@@ -41,6 +41,9 @@ const (
 	SchedulerClusterRoleName    = "storageos:scheduler-extender"
 	SchedulerClusterBindingName = "storageos:scheduler-extender"
 
+	SchedulingClusterRoleName    = "storageos:scheduling-extender"
```
darkowlzz, line 44: `Scheduler` for kube-scheduler and `Scheduling` for the extender could be confusing.
How about `SchedExt` or `SchedulerExt`?

mhmxs: The suggested names are also confusing for me. Which one is which? Let me find some better name for them.

pkg/storageos/rbac.go
```diff
@@ -38,11 +38,11 @@ const (
 	NFSClusterRoleName    = "storageos:nfs-provisioner"
 	NFSClusterBindingName = "storageos:nfs-provisioner"
 
-	SchedulerClusterRoleName    = "storageos:scheduler-extender"
-	SchedulerClusterBindingName = "storageos:scheduler-extender"
+	KubeSchedulerClusterRoleName    = "storageos:kube-scheduler"
+	KubeSchedulerClusterBindingName = "storageos:kube-scheduler"
```
darkowlzz, line 42: This name change is good but I'm afraid we shouldn't change an existing resource name. This will require migration process/step to the new version.

croomes (approved):

PR #340, opened 2021-06-11
Operator shouldn't tolerate NoSchedule taint

croomes:
The uber-yaml sets the Operator pod to tolerate NoSchedule, which it doesn't need to do.  Tolerations for the other components are set in the StorageOSCluster CR and remain unchanged.


avestuk:
LGTM

darkowlzz (approved):
LGTM!

PR #343, opened 2021-07-12
Prep for v2.4.2

croomes:
Prepare for v2.4.2 release.

Update internal/pkg/image/image.go and README.md for new node image.
Update NEW_VERSION ?= "v2.4.2" and CURRENT_BUNDLE_VERSION = "2.4.2" in Makefile
Run NEW_VERSION=v2.4.2 make release
Run make bundle
Remove "replaces" line from bundle/manifests/storageosoperator.clusterserviceversion.yaml and revert removal of resources.
Revert description formatting change in  deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml and deploy/olm/storageos/storageos.clusterserviceversion.yaml


bundle/manifests/storageosoperator.clusterserviceversion.yaml
```diff
@@ -49,190 +49,8 @@ spec:
         name: nfsservers.storageos.com
         version: v1
       - description: StorageOSCluster is the Schema for the storageosclusters API
-        displayName: StorageOS Cluster
         kind: StorageOSCluster
         name: storageosclusters.storageos.com
-        resources:
-        - kind: Namespace
```
croomes, line 56: @darkowlzz I'm not sure why `make bundle` removed this - shall I revert?

deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
```diff
@@ -82,89 +82,55 @@ metadata:
...
-
-    ## About StorageOS
-
-    StorageOS is a software-defined cloud native storage platform
-    delivering persistent storage for Kubernetes. StorageOS is built from
-    the ground-up with no legacy restrictions to give enterprises working
-    with cloud native workloads a scalable storage platform with no
-    compromise on performance, availability or security. For additional
-    information, visit [www.storageos.com](www.storageos.com).
+  description: "StorageOS is a cloud native, software-defined storage platform that\ntransforms
```
mhmxs, line 85: I more like the previous version of this description. Should we use `|` instead of `"`?

bundle/manifests/storageosoperator.clusterserviceversion.yaml
darkowlzz: Yes, this is a known problem in the existing bundle generation, which is based on some old tools. For now, this block has to be added back. The new tooling has much better way to generate and populate this using kustomize and some code markers.

deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
darkowlzz: This seems to be coming from some updated tool, not sure which one, probably yq, although we use a pinned version. I tried previewing this CSV in the https://operatorhub.io/preview and the result looks same. We can manually change it back as this may be the last time we are touching this or keep it as it is.
Again, the new tooling in the new operator using kustomize makes all these better.

mhmxs (approved):

deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
mhmxs: Ok i see, i have approved the change, it is up to @croomes to change it if he agrees.

croomes: I'm not sure why, but the generation tools changed this.

bundle/manifests/storageosoperator.clusterserviceversion.yaml
croomes: thanks - done

deploy/olm/csv-rhel/storageos.clusterserviceversion.yaml
croomes: I've reverted - I prefer the old way too.

PR #345, opened 2021-08-11
Add badges to readme

mhmxs:


darkowlzz:
We are about to deprecate this project in favor of the [new operator](https://github.com/storageos/operator) and it's already in a maintenance mode. I'm fine with merging it, but if we add the project status badge now, it'll need to be changed in a few weeks.

mhmxs:
As i know this is our active operator, and it will still active until the end of lifecycle a.k.a. no-one uses it anymore :) I'm glad to come back and de-activate it.

PR #350, opened 2021-09-07
Generate separated image with manifests

mhmxs:
Operator manifests now generated as a separated docker image for both develop and tagged versions.


.travis.yml
```diff
@@ -11,7 +11,7 @@ env:
     - K8S_VERSION="v1.19.0"
     - MINIKUBE_VERSION="v0.28.2"
     - IMAGE_NAME=storageos/cluster-operator
-    - IMAGE_TAG=test
```
mhmxs, line 14: Previously for `operator-image` target we used `IMAGE_TAG` but for `install-manifest` we used `TRAVIS_TAG` and then we re-tagged image. With this change we build only `TRAVIS_TAG` so w save an extra tagging.

.travis.yml
```diff
@@ -71,8 +71,8 @@ jobs:
             echo "Docker credentials not set. Skipping docker log in...";
           fi
       script:
-        - make operator-image
-        - OPERATOR_IMAGE=storageos/cluster-operator:$TRAVIS_TAG make install-manifest
+        - |
+          (: ${TRAVIS_TAG:=develop} ; OPERATOR_IMAGE=$IMAGE_NAME:$TRAVIS_TAG MANIFESTS_IMG=$MANIFESTS_IMAGE_NAME:cluster-$TRAVIS_TAG make operator-image install-manifest manifests-image)
```
mhmxs, line 75: Set `TRAVIS_TAG` if not setted.

Makefile
```diff
@@ -1,4 +1,6 @@
 OPERATOR_IMAGE ?= storageos/cluster-operator:test
+# Image URL for manifest image.
+MANIFESTS_IMG ?= storageos/operator-manifests:cluster-test
```
croomes, line 3: We use `test` in most places - any reason for `cluster-test`?

scripts/deploy.sh
```diff
@@ -31,6 +32,6 @@ if [ "$1" = "tagged" ]; then
        tianon/github-hub:2 bash -c "./scripts/create-pr.sh"
 
 elif [ "$1" = "develop" ]; then
-    docker tag "${IMAGE_NAME}:${IMAGE_TAG}" "${IMAGE_NAME}:develop"
     docker push "${IMAGE_NAME}:develop"
+    docker push "${MANIFESTS_IMAGE_NAME}:cluster-develop"
```
croomes, line 36: `develop`?

croomes (commented):
A couple minor naming things that might be fine, they just appear inconsistent.

Makefile
mhmxs: New operator uses the same docker repo (as you suggested), and we are not able to use the same tag.

scripts/deploy.sh
mhmxs: https://github.com/storageos/cluster-operator/pull/350#discussion_r703644066

Makefile
mhmxs: If you create a separated repo for cluster operator manifests, i'm glad to use it.

croomes: makes sense - thanks

croomes (approved):

